PR #59, opened 2014-05-21
added working SplitImage experimental Privly App

smcgregor:
This is a proof-of-concept application for sharing images via HTML5 file APIs. It pipes the images through the ZeroBin style of encryption so the image is not stored in cleartext.


smcgregor:
This app is being developed on the [GSOC-develop branch](https://github.com/privly/privly-applications/tree/gsoc-SplitImage) and being tracked by issue #160. I am about to push an update to the branch so it will track with the newer MVC architecture.

PR #77, opened 2014-07-22
Use auth tokens to login

fullvlad:
Users can choose to stay logged in even after the browser is closed. This is achieved using an authentication token requested from the server that is save in localStorage. When the user logs out, the token is removed from the server.


smcgregor:
I substantially increased the session time on the server (maximum 4 weeks and 2 weeks max without activity). I think this mitigates the login UX issue without needing to use the permanently logged in token. Feel free to reopen this if you still want it.

Issue #110, opened 2014-11-15
Remove PlainPost and rename ZeroBin


irdan:
Giving users a choice between PlainPost and ZeroBin serves as a point of confusion.  PlainPost and ZeroBin do not differ significantly from the perspective of users.  We should remove PlainPost, and rename ZeroBin to message.

This eliminates confusion about ZeroBin being a separate peice of software from Privly, and makes the right choice for users by default.  Of course, documentation and threat models will make reference to ZeroBin, but general users should not be exposed this level of detail.

This will simplify rollout initially, and additional applications can be added over time with meaningful-to-user application names.

smcgregor:
This is done.

PR #127, opened 2014-11-16
Rename zerobin

irdan:
This pull request renames the Zerobin App to Message.  It addresses one half of #110.


irdan:
Aside from the show.html in the ZeroBin folder needing to be built, I believe this is ready to go.  I assume that you will be rebuilding the html as part of the merge so won't bother doing that here.

PR #140, opened 2015-01-04
Fix storage test failures

irdan:
The tests were failing in firefox because localStorage["key"] is not the same as localStorage.getItem("key").  This fixes that issue.

This also changes the default behavior of localStorage to explicitly return undefined when a key is undefined. By default localStorage returns null when a key is undefined.

### DIFF
--- a/shared/javascripts/local_storage.js
+++ b/shared/javascripts/local_storage.js
@@ -24,11 +24,12 @@ var ls = {
     }
 
     if ( ls.localStorageDefined ) {
-      return localStorage.setItem(key, value);
+      localStorage.setItem(key, value);
     } else {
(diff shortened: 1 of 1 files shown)


shared/javascripts/local_storage.js
```diff
@@ -24,7 +24,7 @@ var ls = {
     }
 
     if ( ls.localStorageDefined ) {
-      return localStorage[key] = value;
+      return localStorage.setItem(key, value);
```
smcgregor: These have different behavior. The new one does not return the value.

irdan: That is a really good catch.  I'll make the change so that the behavior stays the same.

Issue #109, opened 2014-11-15
The content of the first_run page is confusing


irdan:
Right now when a user installs the extensions they are greeted with a changelog.  This means nothing to most users, and only serves to confuse them.  We should use this opportunity to give meaningful information to new users.  

We should decide as a group precisely what content should live there.  A quick guide integrated with a 'playground' for learning how to use privly comes to mind.  This is not an ideal solution since we shouldn't have to train people to use the software, but it is leaps and bounds ahead of not providing that training.

I would like to hear thoughts on this idea, and other ideas for the content that should live on the first run page.

See also https://github.com/privly/privly-chrome/issues/33

PR #143, opened 2015-01-24
First run page

spectralsun:
#109


smcgregor:
@spectralsun Looks good, I'll write new content.

@irdan Can you record a new GIF with the message "Using Privly means websites won't read your messages." The example GIF should not target a single website and I think more GIFs will clutter the page.

PR #149, opened 2015-02-12
Made some minor style changes according to CodeClimate's recommendations...

katten-kits:
.... Mostly declaring undeclared variables and using regex quotes (/regex/ instead of "regex").

### DIFF
--- a/Pages/js/options.js
+++ b/Pages/js/options.js
@@ -38,7 +38,7 @@ function saveCheckedSetting() {
 
   // Don't run on Firefox
   var status = document.getElementById("button_status");
-  if (status === null ) {return};
+  if (status === null ) {return;}
 
   // reset status
   status.innerHTML = "";
@@ -54,7 +54,7 @@ function saveCheckedSetting() {
  */
 function restoreCheckedSetting() {
   var btn = document.getElementById("disableBtn");
-  if ( btn === null ) {return}; // Don't run on Firefox
+  if ( btn === null ) {return;} // Don't run on Firefox
   var stored = ls.getItem("Options:DissableButton");
   btn.checked = (stored !== undefined && stored === true);
 
@@ -77,7 +77,7 @@ function saveWhitelist() {
   var user_whitelist_input = csv;
   
   // characters to split entered domains on
-  var invalid_chars = new RegExp("[^a-zA-Z0-9\-._]","g"); 
+  var invalid_chars = new RegExp(/[^a-zA-Z0-9\-._]/, "g"); 
   var domains = user_whitelist_input.split(invalid_chars); 
 
   // Each subdomain can be from 1-63 characters and may contain alphanumeric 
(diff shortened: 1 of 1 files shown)


Pages/js/options.js
```diff
@@ -85,19 +85,19 @@ function saveWhitelist() {
...
-  var notEndInHyphen = new RegExp("[^\-]$","g");
+  var notEndInHyphenOrUnder = new RegExp(/[^\-_]$/, "g"); 
+  var notEndInHyphen = new RegExp(/[^\-]$/, "g");
 
   var domain_regexp = "";  //stores regex to match validated domains
   var valid_domains = [];  //stores validated domains
   
   //iterate over entered list, split by invalid chars
-  for (var i = 0; i < domains.length; i++){ 
+  for (i = 0; i < domains.length; i++){
```
smcgregor: I believe you made this global.

smcgregor:
Looks good. If you address the one comment I'll test/merge.

Pages/js/options.js
katten-kits: Wow, thanks for the quick feedback! The reason I changed this is because Javascript doesn't use block scope (only function scope), so this variable was already "global" to the function. (This was the issue with a few other declarations as well.) So, this basically removes a redundant declaration.

smcgregor: > function hi() { aa = "b";}
> hi();
> console.log(aa); // gives "b"

Unless a for loop declaration has special semantics, I still think `i` is now global.

katten-kits: Ohh, I see what you mean: if you don't declare the variable, it is global. However, "var i" was declared earlier in the function (in line 71), so I didn't remove *all* declarations, just a redundant one. It is still local to the function.

Would it be more clear if I moved "var i;" to the top of the function and then used "i = 0" in both for loops? Or is it okay this way?

smcgregor: Ahhhh, k. Maybe declare it before the first use to show it is used more
than one place. At least that is how I would read it.

katten-kits: That's a good idea. I'll change it so it's more clear. Thanks for your help!

smcgregor:
Looks good, I'll test and likely merge it within the week.

smcgregor:
This failed testing since I am unable to update the user whitelist in the options page.

You are halfway between the two valid syntaxes of regular expressions.

1. They should not use the "new" declaration, ie: `new RegExp(/^(?!\-|_)[\\w\-]{1,63}/, "g");`, but instead should be `/^(?!\-|_)[\w\-]{1,63}/g;`. Otherwise there is an error in the console.
2. Remove the "escape" character `\` where appropriate. Doing this will require some [reading](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions). The problem is that the escape characters are not supposed to be in the syntax you changed them to.

katten-kits:
Ah, can't believe I forgot to consider that. Thanks again for the feedback, I apologize for the errors! Will improve those.

katten-kits:
Alright, I went through and changed the syntax all the way. This should work better!

smcgregor:
This looks good. I'll test and will likely merge within the next week.

smcgregor:
All good!

:)

PR #170, opened 2015-02-28
Integration script fix

irdan:
Minor fixes to the `run_all.rb` script.

### DIFF
--- a/test/selenium/run_all.rb
+++ b/test/selenium/run_all.rb
@@ -225,6 +225,11 @@ end
 
 if platform == "sauce_chrome_extension"
 
+  if `pwd`.include? "privly-firefox"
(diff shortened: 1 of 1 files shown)


irdan:
I added the check to the one other point where relative paths are used in the script.

test/selenium/run_all.rb
```diff
@@ -154,15 +154,20 @@
 # Package and add the Firefox extension as necessary
 if platform.include? "firefox_extension"
 
+  if `pwd`.include? "privly-chrome"
+    puts "\nCannot test the firefox extension from within the chrome extension\n\n"
+    exit 0
```
smcgregor: `exit 1` is preferred here since that indicates a fail condition.

irdan:
I updated the exit values.

PR #173, opened 2015-03-01
test injection of reddit.com

spectralsun:


test/selenium/specs/tc_show_injected.rb
```diff
@@ -31,6 +31,19 @@ def test_showing_injected_posts
...
+
+    if not @@privly_extension_active
+      return
+    end
+
+    # Open these pages
+    test_page = "http://www.reddit.com/r/Privly/comments/2xipzl/privly_demonstration_link/"
+
+    visit test_page
+    assert page.has_css?(".usertext-body p iframe")
```
smcgregor: Referencing the css of the page in this way is very brittle. I recommend finding the URL's string and looking for a sibling iframe. This will work even if they change the whole application.

spectralsun:
This has been updated, please review and merge.

PR #179, opened 2015-03-02
JSON Objects communication

bhavul:
The sending of resize message works.


bhavul:
@smcgregor check host_pages_integration.js 
I know I didn't have to add the code I did. But, that's just to show how I've checked the working of all 4 functions as of now. I'll remove them. Just wanted you to have a look at them once. I've commented them out. L39 to L52.

bhavul:
@smcgregor Added other methods too. Only  adding tests remaining now, for this and privly.js file.

shared/javascripts/host_page_integration.js
```diff
@@ -32,8 +32,69 @@ var privlyHostPage = {
...
+     * Send the parent document the name of this iframe. This will
+     * allow the host page to hide the iframe and show the original link
+     * instead. 
+     *
+     */
+    hideWrapper: function(){
+        var evt = document.createEvent("Events");  
+        evt.initEvent("IframeRemoveEvent", true, false);
+        var element = document.createElement("privElement");
+        var frameId = window.name;
```
smcgregor: These functions are largely duplicated 3 times. Maybe one helper function that is called with three different params?

smcgregor:
Are you still working on tests for this?

bhavul:
@smcgregor I'll do the helper function part and yeah, I had started working on the unit tests. I'll add those too in the next 3-4 days.

smcgregor:
Cool. This will be a great feature to have :)

smcgregor:
This is still marked as "Needs Work." Is there something I am waiting on at this point, or should I be reviewing this PR?

bhavul:
Yes, just give me few more hours. I had got stuck at some silly place in integration testing and then have been a bit away due to academics. But, I'll add integration testing(it's almost done) and then leave you another message here to review it.

bhavul:
@smcgregor Just review it once. 

Also, I have a query about integration testing file. As far as I've understood other such ruby and capybara files, they pseudo-act like users. Click on buttons, opens navigation and so on to check how the whole system works.
I've introduced an API here, which will be used by privly app developer. So, there's no such user action right now that can trigger showing/hiding/removing injected links. What should then be included and tested in an integration testing file for this?

smcgregor:
I think you might be limited to unit testing until an app implements the API.

bhavul:
I've added unit testing already. You can then review, I think.
Shoot, i should have then asked you to review this way way before. It's
been weeks since I completed unit testing.

smcgregor:
[I wrote a shared messaging library](https://github.com/privly/privly-applications/blob/gsoc-develop/shared/javascripts/context_messenger.js) that @SummerWish abstracted so each of the scripting contexts could message each other across Jetpack/Chrome/Safari. The library should make it easier to integrate these changes across all platforms.

Sorry it has taken me so long to update you here. This has been noted in my email for 3 months. I think it would be best to test and merge this into develop once the context_messenger.js lands in develop.

PR #181, opened 2015-03-03
Wrap an immediate function invocation in parens to improve code quality

ujjwalwahi:
Improves the code quality of *Message/js/base64.js* by wrapping an immediate function invocation in parens

### DIFF
--- a/Message/js/base64.js
+++ b/Message/js/base64.js
@@ -12,11 +12,11 @@ if (global.Base64) {return;}
 
 var b64chars 
     = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/';
-var b64tab = function(bin){
+var b64tab = (function(bin){
     var t = {};
     for (var i = 0, l = bin.length; i < l; i++) {t[bin.charAt(i)] = i;}
     return t;
-}(b64chars);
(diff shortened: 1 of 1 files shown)


spectralsun:
:+1:

irdan:
Since `base64.js` is part of the ZeroBin library, I'm not sure we want to start modifying it.  Unless we want to either submit this upstream or are openly okay with diverging from the fork, I'm -1 to merging this.

smcgregor:
I didn't put it in the vendor directory, so I think it is fair game. The only issue I can think of is that Mozilla _may_ want to verify its integrity against the source before they will distribute the extension. However, I don't think this is actually an issue here since the ZeroBin author seems to have already modified it.

Regardless, this is a recommendation of CodeClimate? I am surprised it wants the extra parens.

@ujjwalwahi: I am +1 if you link to the recommendation on CodeClimate.

smcgregor:
For reference, this is the build on my fork. https://travis-ci.org/smcgregor/privly-applications/builds/53004176

ujjwalwahi:
Link to CodeClimate recommendation https://codeclimate.com/github/privly/privly-applications/Message/js/base64.js

smcgregor:
Sound reasoning, thanks!

PR #184, opened 2015-03-04
Additional Check for details(Login/Register)

hitesh96db:
If the user clicks on "Login" or "Register" before filling out any detail, a meaningless request is sent to the server. Instead, we could check if the user has filled any details before making a request to the server.
Thoughts on this ?

### DIFF
--- a/Login/js/new.js
+++ b/Login/js/new.js
@@ -41,21 +41,18 @@ var callbacks = {
     // Set the nav bar to the proper domain
     privlyNetworkService.initializeNavigation();
     
-    // Monitor the login button
-    document.querySelector('#login').addEventListener('click', callbacks.submitCredentials);
-    document.querySelector('#register').addEventListener('click', callbacks.submitRegistration);
-    $("#user_password").keyup(function (e) {
-        if (e.keyCode === 13) {
-            callbacks.submitCredentials();
-        }
+    var loginForm = document.getElementById("loginForm");
+    var registerForm = document.getElementById("registerForm");
+
+    loginForm.addEventListener('submit', function(e) {
+      e.preventDefault();
+      callbacks.submitCredentials();
+    });
+    registerForm.addEventListener('submit', function(e) {
+      e.preventDefault();
+      callbacks.submitRegistration();
     });
 
-    $('#register_email').keyup(function (e) {
-        if (e.keyCode === 13) {
-            callbacks.submitRegistration();
-        }
-    });
-    
     // Add listeners to show loading animation while making ajax requests
     $(document).ajaxStart(function() {
       $('#loadingDiv').show(); 
@@ -85,7 +82,7 @@ var callbacks = {
    * Submit the posting form and await the return of the post.
    */
   submitCredentials: function() {
-    $("#save").prop('disabled', true);
+    $("#login").prop('disabled', true);
     privlyNetworkService.sameOriginPostRequest(
       privlyNetworkService.contentServerDomain() + "/users/sign_in", 
       callbacks.checkCredentials,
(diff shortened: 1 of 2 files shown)


Login/js/new.js
```diff
@@ -85,23 +85,35 @@ var callbacks = {
...
-      privlyNetworkService.contentServerDomain() + "/users/sign_in", 
-      callbacks.checkCredentials,
-      {"user[email]":  $("#user_email").val(),
-       "user[password]":  $("#user_password").val()
-       });
+    
+    var userEmail = $("#user_email").val();
+    var userPassword = $("#user_password").val();
+    if (userEmail != "" && userPassword != "") {
+      $("#save").prop('disabled', true);
```
smcgregor: You should refactor this into an event that fires on keydown and submit and checks for contents in the form. I think it is possible to permanently lock the form if they submit the wrong password/email combo, no?

hitesh96db:
@smcgregor Fixed the button 'lock' bug and refactored the code.

smcgregor:
Can you refactor this a bit so that the validation logic is not nearly duplicated? The same could be said of the sameOriginPostRequest code, but I believe I introduced that.

hitesh96db:
@smcgregor What do you think about the refactoring ? If I'm doing good, I could try the same for sameOriginPostRequest ?

smcgregor:
Moving in the right direction, but you can still improve the new code.
Anytime you see a repeated line view it as an opportunity to refactor. This
will also make it easier to simplify the conditional statements.

Keep massaging. :)

hitesh96db:
@smcgregor Done.(I might be running out of ideas now ;))

smcgregor:
It is difficult for me to make suggestions from my phone...

I think perhaps we are overthinking this. Try searching "html5 validators".
I think the newer apis might take care of this.

hitesh96db:
Actually, "required" would do the job as the html validator, which I tried initially, though that would only work if the form has an action, which is called on clicking submit. In the current implementation, event handlers are used to submit the form. However, we would have to add an event handler to notice if the user has pressed "enter".

smcgregor:
You can also capture the form's submission in js. The action could then be
#.

hitesh96db:
@smcgregor I've changed my approach towards this. This has now made things a lot better:
1) "required" checks for empty fields
2) Emails fields are validated for an email.
Firefox and Chrome automatically submit forms on pressing enter, so I've removed the event handler to check for key press.

smcgregor:
Note: In the future you should develop from a feature branch. Otherwise you will run into problems when master diverges from your master.

smcgregor:
For my own notes, I attempted to merge the two registration/login pull requests by I am receiving an error.

[branch](https://github.com/smcgregor/privly-applications/commits/registration-page-update)

[build](https://travis-ci.org/smcgregor/privly-applications/builds/53305951)

hitesh96db:
Can you guide me through this ? What should I do now - write tests ?

smcgregor:
If you want to work on this more, you should make a pull request against my branch containing the two pull requests. It needs an update to the unit tests defined here:

https://github.com/privly/privly-applications/blob/master/Login/test/new.js

You should remove all the references to the HTML2JS code that loads DOM from the HTML file. Instead, you should write code that isolates individual functions and tests their results. This will be a good learning exercise.

hitesh96db:
I'll get started with it :+1:

hitesh96db:
@smcgregor I need some help regarding these tests. I need to know how to write them.
My approach so far --
> call callbacks.pendingLogin
> call callbacks.submitCredentials
> check if callbacks.checkCredentials is called using "spyOn" of Jasmine

However there are two ajax calls taking place to authenticate
1) Get CSRF token
2) Send the credentials
In order for the functions to complete/pass tests, these ajax calls must be completed. However, I am unable to get  them to work with tests, the ajax calls fail during the test.

smcgregor:
You should isolate individual functions by assigning a shim function to functions that are called by the isolated function. For instance, if a function uses an AJAX request, you can reassign the AJAX request to return something like: `{privlyUrl: "http://example.com/test_case"}` instead of actually generating a request on the network.

Issue #157, opened 2015-02-23
Develop with Unit Tests


smcgregor:
Develop Privly using the Karma unit testing system.

This issue is for helping people setup Karma. Comment if the documentation for unit testing is not sufficient and you need help. We will then translate any issues or misconceptions into changes in the documentation.

PR #194, opened 2015-03-05
Add a test for generateNewGlyph in tooltip.js

d356:
Test to see if generateNewGlyph successfully adds glyph_cells and glyph_color to local storage, and that glyph_cells has 15 items. #157


smcgregor:
Looks good. Just need to test.

PR #196, opened 2015-03-05
Karma test, follow FILES_TO_TEST convention

sammyshj:
Follow the karma test convention for setting the FILES_TO_TEST variable. First the file to be tested on and next the actual tests.


smcgregor:
Good call.

Issue #189, opened 2015-03-04
SplitImage Branch Update


smcgregor:
The [experimental branch](https://github.com/privly/privly-applications/tree/experimental-SplitImage), SplitImage, is stale and needs updating. I believe GSoC applicants want to work on this.

See also #160

vatsalj:
@smcgregor : I believe this issue can be closed since #201 is merged.

PR #197, opened 2015-03-05
Update split image branch to include latest commits

anmoljagetia:
in reference to #189.

PR #195, opened 2015-03-05
Separated Login form and Register form

SummerWish:
#163 
And with some minior fixes. See commit logs for details.

login

register

PR #198, opened 2015-03-05
Updated SplitImage branch

SummerWish:
Merged master branch to SplitImage branch. #189 

splitimage_show


SummerWish:
Oh, just had dinner with friends and it's late :-(
Might be duplicate with https://github.com/privly/privly-applications/pull/197

smcgregor:
They are duplicates. Do you see any advantage to working from your PR, or should I just look at the first one to climb the hill?

SummerWish:
Well, I just compared two branches.

- `build.py`
mine commit doesn't made changes to `build.py` compared to current master branch and everything builds well. #197 contains several outdated config entries which I think isn't necessary and might break it. https://github.com/anmoljagetia/privly-applications/blob/Update-SplitImage/build.py#L160

- `base64.js`, `rawdeflate.js`, `rawinflate.js`, `zerobin.js`
I moved these four files to `/shared/ZeroBin` to share with both SplitImage and Message. #197 keeps two copies (`/shared/ZeroBin` and `/Message/js`). (https://github.com/anmoljagetia/privly-applications/tree/Update-SplitImage/shared/ZeroBin and  https://github.com/anmoljagetia/privly-applications/tree/Update-SplitImage/Message/js)

- `shared/test/privly-web/new.js`
I didn't apply any changes to this file (I think it is unrelated to SplitImage after a quick review). #197 merges it. BTW #197 has some unsolved merge conflicts (https://github.com/anmoljagetia/privly-applications/blob/Update-SplitImage/shared/test/privly-web/new.js#L74).

- `zerobin.js`
I used latest `zerobin.js` (which is located in `Message/js` directory) in `shared/ZeroBin`. #197 uses a former version here.
https://github.com/anmoljagetia/privly-applications/blob/Update-SplitImage/shared/ZeroBin/zerobin.js#L26 vs new version https://github.com/SummerWish/privly-applications/blob/experimental-SplitImage/shared/ZeroBin/zerobin.js#L26

- `ZeroBin/*`
I didn't apply changes to files in this directory since it is deprecated. #197 applies. 
https://github.com/anmoljagetia/privly-applications/blob/Update-SplitImage/ZeroBin/new.html.subtemplate#L8

Other differences are mainly caused by `build.py`: I ran `build.py` to rebuild all templates, which inserts SplitImage entry to navigation bar.

However I made a mistake here :-( my SplitImage branch is based on my previous work (https://github.com/privly/privly-applications/pull/195), which should not be included in this branch. I will fix it later.

smcgregor:
Thanks for the summary.

It would be better to leave base64.js, rawdeflate.js, rawinflate.js, and zerobin.js in the Message directory for the time being. We have modified them, but they are not our libraries so they don't exactly fit in either the vendor or the shared directories. Let's just reference them across apps for the time being. This isn't good style, but I would like to evaluate the more recent options for JS compression before we adopt them as the recommended standard -- particularly for SplitImage.

smcgregor:
Closed for https://github.com/privly/privly-applications/pull/201

Thanks for addressing the comments, but these became very difficult to review.

Issue #158, opened 2015-02-23
Add Integration Tests


smcgregor:
Develop Privly's integration tests that use Selenium WebDriver.

This issue is for helping people setup Selenium WebDriver. Comment if the documentation for Selenium WebDriver is not sufficient and you need help. We will then translate any issues or misconceptions into changes in the documentation.

PR #200, opened 2015-03-05
Add integration test for the history page

d356:
Test if visiting the history page and clicking "Preview Message" works and brings up the modal dialog. Also test closing the dialog.  This test passes for me locally as well as with Sauce Labs. #158


smcgregor:
Code looks good.

Issue #166, opened 2015-02-28
Change Posting Prompts to be a Clearer Step by Step


spectralsun:
When you are posting a new link without originating it from the context menu (right click) or the posting button, you see a series of messages that tell you what to do. This is essentially a step-by-step guide, but it is currently displayed as a series of boxes that appear after clicking in the proper place. The step-by-step nature of this process could be made more explicit by perhaps adding numbers to the left hand side of the popup.

smcgregor:
Screen cap for reference (see the yellow area). 
![screen shot 2015-02-27 at 6 12 34 pm](https://cloud.githubusercontent.com/assets/64780/6424355/6155d9be-beac-11e4-8910-41a4fb184c03.png)

spectralsun:
Pull #192 looks good except it seems to me that step 1 and 2 are the same, could we get rid of step 1 and the word "Now" on step 2 for simplification? Having 3 messages is kind of overwhelming.

d356:
@spectralsun Thanks for the feedback. Do you mean like this?
![privly-2](https://cloud.githubusercontent.com/assets/10871998/6493235/5bcf6b0c-c287-11e4-9d72-53192064d6e7.PNG)

smcgregor:
Ha, I don't think he did _exactly_ but I was just about to suggest this.

spectralsun:
@d356 I was thinking just remove the top line all together, just have the https link (as the first message) and the message below about copy/paste and the button top open.
@smcgregor thoughts?

smcgregor:
Based on Think-Aloud study results, we must tell the user what to do with the link. So the first message is required.

vatsalj:
Hi privly devs! I was looking into this issue and I feel that the link in a separate `<p>` tag makes it look like an additional step. I made something like this:
![privly snapshot](https://cloud.githubusercontent.com/assets/1089034/6525498/fb90b45c-c42b-11e4-8b41-d5333a3322e4.png)

The advantage of this can be that it maintains the continuity of Step 1 while separating the steps into distinctive paragraphs. Furthermore, the instructions and action-to-be-taken part of 1st step are separated by a horizontal rule as well. Do let me know if this looks like a viable solution to you.

smcgregor:
Looks good to me. @spectralsun, how does it look to you? I think I would make a few changes, but they are more tweaks like rounding the corners, playing with the width of the horizontal rule, etc.

vatsalj:
[...]

PR #202, opened 2015-03-06
Issue #166 : Change Posting Prompts to be a Clearer Step by Step

vatsalj:
There are inline styles in the `<hr>` tag, those can be removed if required. Also, text for step 1 can be changed if any of the devs think its needed.


vatsalj:
I removed the `<hr>` tag as the markup wasn't w3c compliant. Instead, I used styling to separate text from link.

Issue #206, opened 2015-03-09
Obsolete chrome/firefox-extension-master?


SummerWish:
Since the difference between `chrome/firefox-extension-master` and `master` is only generated HTML, I think it is better to separate development and deployment explicitly, that is ignoring all changes made by `build.py` and then maintaining `master` branch only. (move build phase from development to deployment).

It is easy to build specific platform from source code when deploying:
```bash
git checkout .
python build.py --platform=chrome
# command to pack chrome extension here
```

Advantages:
- maintain source code only
- no more commits of generated HTML
- maintain one master branch, no more merge conflicts
- it's a *common pattern*, for example [jQuery](https://github.com/jquery/jquery/blob/master/.gitignore#L11) repo ignores all generated files in build phase

Disadvantages:
- cannot easily diff built files (if we can ensure the correctness of source code and building step, I think this isn't a big issue)

smcgregor:
I think we are heading in this direction, but we won't make this transition until the next lull in development.

Disadvantages:

* We would need to run the build script on TravisCI for privly-chrome, privly-firefox, etc.
* It adds a step when people clone the project for the first time since `git clone --recursive` no longer gets rendered HTML.
* (maybe) To prevent people from failing to render the HTML, we would probably want to put a placeholder HTML page that tells them they still need to render the HTML.

SummerWish:
Agree, maybe after GSoC LOL.

smcgregor:
Summarizing my thoughts here. We need to change the way HTML is maintained in our code base because of the following considerations:

1. We don't update the HTML every time we update the templates or pages because it clutters the commit logs.
2. It is useful to view the diffs of HTML periodically to check for build issues without visually inspecting all HTML files.
3. It is annoying to have dirty HTML files locally that won't be committed by anyone but me (the person building HTML for the project)
4. The .travis.yml file does not build the HTML, which can break the build in some cases.
5. When new developers clone a server or extension they think it should work without building HTML.

Each of these considerations produces different recommendations, so it is important to find a proper balance.

1. For point (1), the solution is to either keep the file out of the git repo and add it to the .gitignore, or to explicitly ignore it locally by issuing `git update-index --assume-unchanged path/to/file`.
2. I should include the files in the source tree so I can easily track the differences in HTML versions through time. This rules out the first solution for point (1), so it is best to use the `--assume-unchanged` flag.
3. The `--assume-unchanged` flag solves this.
4. I'll add building HTML to the .travis.yml files.
[...]

PR #207, opened 2015-03-09
ignore built files

SummerWish:
#206


.gitignore
```diff
@@ -4,3 +4,15 @@
 *~
 test/node_modules/*
 test/coverage
+Help/new.html
+Help/content_server.html
+History/new.html
+Index/new.html
+Login/new.html
```
smcgregor: */*.html would be better than listing all these explicitly?

smcgregor:
1. This pull request establishes the _requirement_ that people run the build system before they can contribute to the project.
2. People will not easily check the changes they make to built HTML by looking at the diff. This could hide errors.

Still, this is a change we need to make. I just need to think about the two points I raise above.

.gitignore
SummerWish: There may be pages not built from the template but ends with `.html` in the future?

smcgregor: Tomorrow's problem.

SummerWish:
Hello, what's your idea of merge this PR?

Currently I have to repeatly reset and remove my built HTML files if I want to switch my privly-application branch into another, which is not very convenient..Lazy is virtue for programmers :-)

smcgregor:
This pull request requires me to write documentation in ~4 different repositories. You might say I am being very virtuous. If this is a high priority for you I can move it higher in the PR queue, but I am currently allocating my time where I think it is most needed

SummerWish:
You are right. You should put efforts on more important things :-) I can bear this little inconvenience

smcgregor:
See #206

PR #210, opened 2015-03-10
Remove old bootstrap styles

SummerWish:
A little clean up to make page layouts behavior more compatible with latest Bootstrap documentation.


smcgregor:
How thoroughly have you tested this? I am prioritizing my PR queue processing and this one will take time to confirm there are no unintended consequences.

SummerWish:
At least login and settings page is OK. I didn't found significant layout issues currently. In fact this commit remove inconformity between the project and bootstrap.

smcgregor:
I think this PR increased the effective line height? There is some extra whitespace at the top of the injected view.

SummerWish:
@smcgregor OK, in injected view, `bootstrap.css` is not included, so the margin of `<p>` becomes bigger.

How about creating a stylesheet specifically for injected view (not for all), includes basic typography styles? Typography styles can be grabbed from bootstrap or any other suitable libraries.

smcgregor:
This can be added here: https://github.com/privly/privly-applications/blob/master/shared/css/injected/injected.css

Do you have a suggested stylesheet?

Issue #191, opened 2015-03-04
Add Unit Testing to SplitImage


smcgregor:
The existing proof-of-concept for SplitImage does not have unit tests. We should add them.

See also #160 #189 #190

vatsalj:
@smcgregor Hi! I was travelling so could not put in another pull request. I remember your comments on the last PR and I would like to clarify a bit about writing the unit test. 
The docs say that ideally, we should create fixtures, otherwise, elements can be created inside the JS itself. So, I would like to ask if in this case, fixture is just an HTML file (or a generic hard-coded HTML code inside some common JS) which I can create? If not that, can you please help me understand what a fixture can be in this case? However, if you recommend not using fixtures at this moment, should I simply create the elements required in the JS itself as follows:

```javascript
var test_div = document.createElement('div');
test_div.addClass("test-div");
document.body.appendChild(test_div);
```

and initialize them using proper initialization functions (such as `privlyNetworkService.initializeNavigation`)?
Please let me know which approach should I take. Or if I'm missing something I request you to give me some more guidance. Thanks!

smcgregor:
Both processes you are describing are creating fixtures. Fixtures are an expression of the pre-conditions, not a specific filetype. We may eventually start defining fixtures in a collection like on Rails (and others), but for now your manual method of creating elements is sufficient.

> ...a test fixture is all the things that must be in place in order to run a test and expect a particular outcome.
> Frequently fixtures are created by handling setUp() and tearDown() events of the unit testing framework. In setUp() one would create the expected state for the test, and in tearDown() it would clean up what had been set up.

smcgregor:
SplitImage has useful code for those looking to implement an image encryption app on top of Privly, but the architecture has some major changes introduced in the gsoc-development branch. It would be better to start over from that branch's Message application.

PR #211, opened 2015-03-10
Issue #191: SplitImage basic unit tests

vatsalj:
I've not added unit tests for checking keys in localStorage yet (as written in Messages tests). I've done so because keys are saved to localStorage in Messages but not in SplitImage. So do you want me to add the feature to save keys in localStorage? Tests for the same can be written then.


vatsalj:
Issue #191 : I've not built the templates now (as they were not built by @smcgregor initially). Do let me know if that's required.

SplitImage/test/new.js
```diff
@@ -5,9 +5,43 @@
  
 describe ("SplitImage New Suite", function() {
   
+  // Get an HTML document defined by the pre-processor.
+  // This is a rough hack because HTML2JS seems to assign the
+  // key to the absolute URL, which is not reliable on
+  // continuous integration.
+  beforeEach(function() {
+    var keys = Object.keys(__html__);
```
smcgregor: You can't use this hack anymore. You must create the DOM elements you need to test inside the javascript.

SplitImage/test/show.js
```diff
@@ -1,13 +1,43 @@
...
-describe ("SplitImaage Show Suite", function() {
-  
+describe ("SplitImage Show Suite", function() {
+
+  // Get an HTML document defined by the pre-processor.
+  // This is a rough hack because HTML2JS seems to assign the
+  // key to the absolute URL, which is not reliable on
+  // continuous integration.
+  beforeEach(function() {
+    var keys = Object.keys(__html__);
```
smcgregor: Don't use the HTML creation hack. Create the DOM nodes in JS.

smcgregor:
Can you please address the comments I left? Also, this PR includes updating the branch from master. These should be two separate pull requests.

vatsalj:
Whoops! I did some mistake while rebasing the branch. I'll try to resolve this. Sorry!

vatsalj:
I think I'll have to start over with my repository. This is getting too complicated. :-/ Thankfully, I always make a backup so work is not lost, yet.

PR #214, opened 2015-03-12
Issue #191: Unit tests for SplitImage.

vatsalj:
Unit test for `initializeApplication` of SplitImage. Used fixtures for creating the initial expected state for the tests.

Changes done in the non-testing scripts:
1. Moved `initializeApplication` inside `DOMContentLoaded` event callback for SplitImage/js/new.js.
2. Created `initializeApplication` to take care of all initialization tasks for SplitImage/js/show.js . Also, kept the same inside `DOMContentLoaded` event callback.

### DIFF
--- a/SplitImage/test/new.js
+++ b/SplitImage/test/new.js
@@ -3,11 +3,45 @@
  * This spec is managed by the Jasmine testing library.
  **/
  
+
 describe ("SplitImage New Suite", function() {
   
-  it("initializes properly", function() {
-    // pass
-  });
-  
-});
+  var new_fixtures = [
+    '<a class="home_domain">Privly</a>',
+    '<a class="legal_nav">Legal</a>',
+    '<a class="account_url">Account<a/>',
+    '<a id="logout_link">Logout</a>',
+    '<ul><li class="mobile_hide"></li></ul>',
+    '<div id="drop_zone"></div>',
+    '<input type="file" id="files" multiple />',
+    '<button id="save"></button>'
+  ]
 
+  beforeEach(function() {
+    for (var i=0; i<new_fixtures.length; i++) {
+      document.body.innerHTML += new_fixtures[i];
+    }
+  });
+
+  afterEach(function() {
+    document.body.innerHTML = '';
+  });
+
+  it("initializes properly", function() {
+    initializeApplication();
+
+    var domain = privlyNetworkService.contentServerDomain();
+
+    if( privlyNetworkService.platformName() !== "HOSTED" ) {
+
+      // if the app is not hosted, the user should first be directed to the
+      // content_server page of the app bundle.
+      expect($(".home_domain").attr("href")).toBe('../Help/content_server.html');
+    } else {
+
+      // if the application is hosted, the URL should connect to the domain's root
+      expect($(".home_domain").attr("href")).toBe("http://" + window.location.href.split("/")[2]);
+    }
+    expect(domain.split("/")[2]).toBe($(".home_domain").text());
+  });
+});
--- a/test/run_each.sh
+++ b/test/run_each.sh
@@ -39,6 +39,8 @@ runTest 'shared/javascripts/privly-web/new.js,PlainPost/js/new.js,PlainPost/test
 runTest 'shared/javascripts/privly-web/show.js,PlainPost/js/show.js,PlainPost/test/show.js'
 runTest 'shared/javascripts/privly-web/new.js,Message/js/base64.js,Message/js/rawdeflate.js,Message/js/rawinflate.js,Message/js/zerobin.js,Message/js/new.js,Message/test/new.js,Message/test/zerobin.js'
 runTest 'shared/javascripts/privly-web/show.js,Message/js/base64.js,Message/js/rawdeflate.js,Message/js/rawinflate.js,Message/js/zerobin.js,Message/js/show.js,Message/test/show.js,Message/test/zerobin.js'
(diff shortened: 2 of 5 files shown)


test/package.json
```diff
@@ -12,11 +12,11 @@
   "devDependencies": {
     "karma": "^0.12.31",
     "karma-chrome-launcher": "^0.1.7",
+    "karma-coverage": "~0.2",
```
smcgregor: It looks like you accidentally included these changes.

test/run_each.sh
```diff
@@ -39,6 +39,8 @@ runTest 'shared/javascripts/privly-web/new.js,PlainPost/js/new.js,PlainPost/test
 runTest 'shared/javascripts/privly-web/show.js,PlainPost/js/show.js,PlainPost/test/show.js'
 runTest 'shared/javascripts/privly-web/new.js,Message/js/base64.js,Message/js/rawdeflate.js,Message/js/rawinflate.js,Message/js/zerobin.js,Message/js/new.js,Message/test/new.js,Message/test/zerobin.js'
 runTest 'shared/javascripts/privly-web/show.js,Message/js/base64.js,Message/js/rawdeflate.js,Message/js/rawinflate.js,Message/js/zerobin.js,Message/js/show.js,Message/test/show.js,Message/test/zerobin.js'
+runTest 'shared/javascripts/privly-web/new.js,SplitImage/js/file_processing.js,SplitImage/js/new.js,SplitImage/test/new.js'
+runTest 'shared/javascripts/privly-web/show.js,Message/js/base64.js,Message/js/rawdeflate.js,Message/js/rawinflate.js,Message/js/zerobin.js,SplitImage/js/file_processing.js,SplitImage/js/show.js,SplitImage/test/show.js'
```
smcgregor: Duplicated.

SplitImage/test/new.js
```diff
@@ -3,11 +3,47 @@
...
+  });
+
+  afterEach(function() {
+    document.body.innerHTML = '';
+  });
+
   it("initializes properly", function() {
-    // pass
+    // todo, this should be called automatically but I don't believe
+    // the DOMContentLoaded event is being fired in this testing environment
```
smcgregor: nothing should happen automatically before running unit tests, since the unit test sets up the pre-conditions of the test. Here you should manually create the pre-conditions of the function without calling initializeApplication.

vatsalj:
@smcgregor I ran the tests using the FILES_TO_TEST method and tests passed for both new.js and show.js. However, when I used run_each.js, it gave me error saying:

Uncaught TypeError: Cannot read property 'offsetHeight' of null
  at /home/vj/.virtualenvs/gsoc/src/privly-applications/shared/javascripts/host_page_integration.js:9

offsetHeight is used only once in host_page_integration.js and that too for `privlyHeightWrapper`. I added it to the fixture, but still the test wouldn't pass with run_each.js. Can you please tell me what am I missing here?

smcgregor:
I think this problem may be taken care of when you address my comments above.

test/package.json
vatsalj: Yes, I'll remove these changes.

SplitImage/test/new.js
vatsalj: Okay, just to clarify, I should remove the `initializeApplication` function call from here and make sure the DOMContentLoaded event is fired? Can I fire that event manually, inside this test itself?

smcgregor: Whoops, you are actually testing `initializeApplication()`. This is fine. The todo is incorrect though, since no events are automatically fired in the testing framework -- at least they should not be.

vatsalj: Okay. Great! I tried firing the DOMContentLoaded event manually here, just to test, and tests started failing (obviously because karma local URL was not same as home_domain url. As for the TODO, I had copied over code from Message but forgot to modify the comments. Will do this now.

test/run_each.sh
vatsalj: @smcgregor I used the JS files inside Messages itself, as same was done in the HTML subtemplates. So in case of show, I'm using the JS inside Messages itself. Do you want me to copy them inside SplitImage?

smcgregor: Ah, I am surprised you didn't need to include the Message libraries in both test sets...

In the future, please include a summary of your pull request. "Unit tests for SplitImage" is not enough information to set expectations when reading through a pull request. What are the unit tests covering? What are they not covering? Do they all pass all the time? What changes to the non-testing scripts were necessary to support testing?

vatsalj: Right, apologies for not adding a summary to the pull request. I'll keep that in mind for the future. As for the Messages library, I think since these tests just take care of initializing the application, there was no need for the Messages library for new.js tests.

vatsalj:
@smcgregor I noticed that Messages has also used localStorage for storing random keys. I think SplitImage should also have a similar workflow. What's your opinion?

smcgregor:
Yes, but let's wait until the current version's unit tests are merged.

vatsalj:
@smcgregor Okay. I've added a comment on the top about the Pull request so that others can understand it clearly as well. Also corrected the issues that were pointed out. Does this pull request seem okay now?

smcgregor:
It needs more review since my comments were mostly for quick feedback. You are in the queue now.

vatsalj:
Okay. I'll start working on the Integration tests then. :)

smcgregor:
Running the build here, if it passes, I will merge: https://travis-ci.org/smcgregor/privly-applications/builds/56256992

PR #223, opened 2015-03-21
Updated Tooltip.js ( Level 0) **Beginner**

jatindhankhar:
Made changes to improve code quality

Merged the multiple breaked lines
Declared the variables inside the loop and did strict comparison inside conditional statements
Also initialized and defined two variables inside of creating and then defining them

### DIFF
--- a/shared/javascripts/tooltip.js
+++ b/shared/javascripts/tooltip.js
@@ -58,9 +58,7 @@ var privlyTooltip = {
 
       privlyTooltip.tooltipMessage = privlyTooltip.appName
         + ":\u00A0" // non-breaking space
-        + newMessage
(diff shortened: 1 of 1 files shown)


smcgregor:
Looks good at the code level. I need to run test it then will likely merge.

jatindhankhar:
Waiting for the review report :+1:

smcgregor:
It is the queue. Given the explosion of pull requests in the last two weeks, you should not expect this to be merged in the next few weeks. I reviewed it on the code level and my only concern is that you may have broken the spoofing glyph by changing the line from `==` to `===`. I don't think this is the case, but I need to fully test it before merging.

Issue #165, opened 2015-02-28
Change "Experts Only" Heading on Help Page


smcgregor:
In Think-aloud studies people view the messaging found below as an indication that it is more secure. Since this is not the case, we should change it to a different heading, maybe "Indie Hosts Only" or "Self-hosts only".

![screen shot 2015-02-27 at 4 39 45 pm](https://cloud.githubusercontent.com/assets/64780/6423739/4c70d63c-be9f-11e4-81ca-157739ba5b6e.png)

smcgregor:
Fixed in https://github.com/privly/privly-applications/commit/7c0c875e8ec3a12db5ea4531de4b10a30990b4e1

PR #224, opened 2015-03-23
Update Sub Template for New Help

jatindhankhar:
As a response to the [Issue #165 ](https://github.com/privly/privly-applications/issues/165)
Replaced the "Experts only" heading with "Indie Hosts only" 

Also replaced the http protocol with https, so as to avoid redirection done by server side and pushing the world towards HTTPS era :smile:

### DIFF
--- a/Help/new.html.subtemplate
+++ b/Help/new.html.subtemplate
@@ -66,8 +66,8 @@
         <span class="glyphicon glyphicon-eye-open"></span> 
         Indie Hosts Only:
       </strong> 
-      Other applications built on Privly provide privacy only in special 
-      circumstances and should be viewed as "experts only."
+      "These applications provide privacy if you host your own content server. 
+      If you don't know what that means, you should not use these apps."
     </p>
     <div class="col-md-3">
       <form action="../PlainPost/new.html" style="display:inline;">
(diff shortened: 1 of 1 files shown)


smcgregor:
Looks good.

One more change would be for the text description: "These applications provide privacy if you host your own content server. If you don't know what that means, you should not use these apps."

jatindhankhar:
Changed the text as per the suggestion [f6a95e4d....] (https://github.com/jatindhankhar/privly-applications/commit/f6a95e4d6b94b70f6d29fa7252f1e6e061110e7a)

Help/new.html.subtemplate
```diff
@@ -64,10 +64,10 @@
     <p>
       <strong>
         <span class="glyphicon glyphicon-eye-open"></span> 
-        Experts Only:
+        Indie Hosts Only:
       </strong> 
-      Other applications built on Privly provide privacy only in special 
-      circumstances and should be viewed as "experts only."
+      "These applications provide privacy if you host your own content server.
```
smcgregor: I didn't intend for you to copy the quotes here.

jatindhankhar: So should I add another real quote or leave it ?

smcgregor: None of the other descriptions have quotes.

smcgregor:
Fixed in https://github.com/privly/privly-applications/commit/7c0c875e8ec3a12db5ea4531de4b10a30990b4e1

PR #227, opened 2015-03-25
Fixes styling(Level 0)

ManrajGrover:
Fixes styling as suggested by Code Climate(Level 0)


smcgregor:
Looks good.

I think there may be another pull request farther up in the queue for this fix, so it might not be needed by the time I get here.

ManrajGrover:
@smcgregor sir,

Sure sir.

PR #233, opened 2015-04-09
Extract common template to a base template

SummerWish:
This allows developers to write a page other than 'show' or 'new' but uses privly fundamental libraries:

```twig
{% extends "templates/base.html.template" %}
```

More specifically, I want to write a callback page of login in seemless-posting. The current code repo only allows redirecting to URLs when login (not calling back JavaScript functions) and I don't want to modify too much code.

### DIFF
--- a/templates/base.html.template
+++ b/templates/base.html.template
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
+
+    <!--
+      This was likely rendered from the Privly-application templating system
+      See: https://github.com/privly/privly-applications/pull/19
+     -->
+
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+    
+    <link href="../shared/css/common.css" media="screen" rel="stylesheet" type="text/css"/>
+    
+    <script type="text/javascript" src="../shared/javascripts/meta_loader.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/parameters.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/network_service.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/local_storage.js"></script>
+
+    {% block head %}{% endblock %}
+  </head>
+  {% block body %}{% endblock %}
+</html>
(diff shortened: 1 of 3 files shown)


templates/base.html.template
```diff
@@ -0,0 +1,23 @@
...
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
+
+    <!--
+      This was likely rendered from the Privly-application templating system
+      See: https://github.com/privly/privly-applications/pull/19
+     -->
+
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+
```
smcgregor: nit: extra spaces

templates/base.html.template
```diff
@@ -0,0 +1,23 @@
...
+    <!--
+      This was likely rendered from the Privly-application templating system
+      See: https://github.com/privly/privly-applications/pull/19
+     -->
+
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+    
+    <link href="../shared/css/common.css" media="screen" rel="stylesheet" type="text/css"/>
+
```
smcgregor: nit: there are extra space characters here

templates/base.html.template
```diff
@@ -0,0 +1,23 @@
...
+    <link href="../shared/css/common.css" media="screen" rel="stylesheet" type="text/css"/>
+    
+    <script type="text/javascript" src="../shared/javascripts/meta_loader.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/parameters.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/network_service.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/local_storage.js"></script>
+
+    {% block head %}{% endblock %}
+  </head>
+  {% block body %}{% endblock %}
```
smcgregor: The head opening/closing tags are in the base template, but the body tags are not. I think it may be more flexible to include the body tag here since it will allow for the base template to include elements in the body without changing the extending template. Hard to say. I think I need to spend time playing with this.

SummerWish:
Hello, could you merge this PR into develop? It cause many conflicts because my embeded-posting is already based on it :-)

smcgregor:
I'm taking care of this now, sorry for the delay.

smcgregor:
[done](https://github.com/privly/privly-applications/commit/53439e59fe20acabc18cf72c7936bf4a4fda286f)

SummerWish:
Thanks!

PR #238, opened 2015-04-23
Removing html2js hack

smcgregor:
This is a work in progress pull request aimed at removing all the old html2js hacks from the tests since all those testing cases should be covered by selenium. I will merge this PR once all the html2js scourge is banished.


smcgregor:
Merged into [develop](https://github.com/privly/privly-applications/tree/develop).

PR #240, opened 2015-05-05
Fix typo & remove unused code in history application

SummerWish:


History/js/new.js
```diff
@@ -267,15 +267,6 @@ function resizeIframePostedMessage(e) {
   }
 }
 
-/**
- * Sends the currently displayed URL to the extension or mobile framework
- * running the applicaiton so it can be submitted to a host page webform.
- */
-function postUrl() {
-  privlyExtension.firePrivlyURLEvent(
```
smcgregor: This case is defunct, but the idea was that someone could post an existing link from the History application. I am not opposed to removing it, but I'd like to point this out before I merge.

smcgregor:
[merged into develop](https://github.com/privly/privly-applications/commit/e9911cbb6e14bcb66628346cc32b162dca93a46b)

PR #242, opened 2015-05-07
Allow reuse of 'ls' shim

hitesh96db:
These changes are necessary so that the "ls" shim can be re-used in Jetpack, as mentioned [here](https://github.com/privly/privly-jetpack/pull/3)

### DIFF
--- a/shared/javascripts/local_storage.js
+++ b/shared/javascripts/local_storage.js
@@ -26,7 +26,7 @@ var ls = {
     if ( ls.localStorageDefined ) {
       localStorage.setItem(key, value);
     } else {
-      ls.preferences.setCharPref(key, value);
+      ls.storage[key] = value;
     }
 
     return value;
@@ -55,10 +55,10 @@ var ls = {
     } else {
       try {
         try { // try to parse stored value as JSON
-          var value = JSON.parse(ls.preferences.getCharPref(key));
+          var value = JSON.parse(ls.storage[key]);
           return value;
         } catch(e){
-          return ls.preferences.getCharPref(key);
+          return ls.storage[key];
         }
       } catch(e) {
         console.warn("Local Storage key was not in storage");
@@ -78,7 +78,7 @@ var ls = {
       return localStorage.removeItem(key);
     } else {
       try {
-        ls.preferences.clearUserPref(key);
+        delete ls.storage[key];
       } catch(e) {
         console.warn("Local Storage key was not in storage when it was removed");
       }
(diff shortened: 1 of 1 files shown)


irdan:
I think @smcgregor should push a new branch called `jetpack` to `privly-applications` so that a PR with jetpack specific changes can target that branch instead of master.  Thoughts?

hitesh96db:
Yeah, I'm fine with that, whatever seems convenient :smile:

smcgregor:
I still need to establish the master/develop/feature branch convention for all the repositories. I would prefer to merge JetPack specific changes into a develop branch (or master) unless it becomes necessary to drop support for the current Xul extension.

irdan:
@smcgregor Since simple-storage is not available to Xul extensions, it seems like a develop branch needs to be created for JetPack specific development.

smcgregor:
@hitesh96db Would you like your own branch of privly-applications for JetPack development or would you like to add JetPack as a distinct platform within privly-applications? The advantage of making it a distinct platform is you don't need to maintain your own development branch of privly-applications.

hitesh96db:
@smcgregor I didn't get what you meant by "adding Jetpack as a distinct platform within privly-applications".

smcgregor:
[This](https://github.com/privly/privly-applications/blob/master/shared/javascripts/network_service.js#L91) is really used in the local storage shim because it only needs to differentiate between platforms that don't have local storage (Xul) and those that have local storage (everything else). Basically instead of ripping out all the Xul support, you can think of JetPack as a different extension and test for it in privly-applications. In most of privly-applications I would prefer to deprecate Xul rather than rip it out, since you would only need to test for the presence of the JetPack APIs. However, I don't want to be overly burdensome with legacy concerns as you and @irdan work on this.

smcgregor:
(updated comment because I left off the "not")

irdan:
I don't see a reason to keep support for XUL alongside JetPack in the code base.  @smcgregor can you help me understand the reasoning?

It seems like it would add a lot of dead code / complexity long term.  Is this a concern that could be addressed with a final XUL github release?

Is there some use case I am not thinking about? Perhaps something on the content server side uses it, but even if that is the case would it really impact users? A content server wouldn't have access to any extension level api so the behavior should in theory be unchanged.

I'm struggling to find a scenario where the answer to a problem is any worse than telling people to upgrade to a more featured and better looking extension.

smcgregor:
The only point where this will be an issue is with storage. All the other differences are expressed in the extension's repository where we won't need to worry about maintaining compatibility with Xul. So we can either have ~20 lines of changes to support both in storage, or we can maintain two separate development branches of privly-applications while we wait to fully deprecate Xul. Not having two branches saves me management time and it saves you and Hitesh the need of rebasing in the future. If you are planning more substantive changes than adding a storage adapter, then we should probably just open a new feature branch...

irdan:
Ah, well this is a much smaller issue then.  I lost sight of how few changes would actually make their way into `privly-applications`. Provided that XUL support in the the shim is removed once JetPack usurps XUL, I'm happy with whatever @hitesh96db thinks is easiest.

hitesh96db:
I'd go with a new Jetpack development branch.

smcgregor:
I pushed a new branch for this purpose, gsoc-jetpack. https://github.com/privly/privly-applications/tree/gsoc-jetpack

hitesh96db:
PR sent to `gsoc-jetpack` branch - https://github.com/privly/privly-applications/pull/244

PR #244, opened 2015-05-13
Allow reuse of 'ls' shim

hitesh96db:


shared/javascripts/local_storage.js
```diff
@@ -26,7 +26,7 @@ var ls = {
     if ( ls.localStorageDefined ) {
       localStorage.setItem(key, value);
     } else {
-      ls.preferences.setCharPref(key, value);
+      ls.storage[key] = value;
```
smcgregor: Does ls have the `setItem` function as well? I'd like to more closely mirror the two if possible

shared/javascripts/local_storage.js
```diff
@@ -87,14 +87,20 @@ var ls = {
   }
 };
 
-// Determine whether localstorage can be used directly
-try { 
-  localStorage;
+try {
+  // check for jetpack runtime environment
+  exports.ls = ls;
```
smcgregor: How does this check for JetPack? What error will it throw?

shared/javascripts/local_storage.js
hitesh96db: Yes, `ls` still remains the same, setItem and getItem don't change.

shared/javascripts/local_storage.js
hitesh96db: exports is not defined.
[exports](https://developer.mozilla.org/en-US/Add-ons/SDK/Guides/Module_structure_of_the_SDK) is a special variable available that allows me to reuse this module in Jetpack.
If this script is executed in a normal/privileged js runtime(i.e, from privly applications), then this would be undefined, though if someone defines exports, it could cause problems.

hitesh96db: I should change it to `module.exports`(CommonJS) and I think the comment is a bit misleading(I really need to just import it as a commonjs module, but that doesn't really check for Jetpack runtime environment).

shared/javascripts/local_storage.js
smcgregor: if you could change this to ls.setItem(key, value), then they will more closely mirror each other. Assignment and function calls can do different things at times so it is better to eliminate that difference between then two branches of the conditionals.

shared/javascripts/local_storage.js
```diff
@@ -87,14 +87,21 @@ var ls = {
...
-try { 
-  localStorage;
-} catch(e) {
-  ls.localStorageDefined = false;
-  
-  // Assuming Xul firefox
-  ls.preferences = Components.classes["@mozilla.org/preferences-service;1"]
-                         .getService(Components.interfaces.nsIPrefService)
-                         .getBranch("extensions.privly.");
+var isCommonJS = typeof window == "undefined" && typeof exports == "object";
```
smcgregor: You should use strict equality checks `===` in almost every case with JavaScript.

smcgregor:
You had a comment that was emailed to me but I think you since deleted or edited. To confirm, this API is accessible to privly-applications when they are injected into the context of a host page? You were making a distinction between jetpack and non-jetpack that indicated this might not be the case.

shared/javascripts/local_storage.js
hitesh96db: Sorry, I misunderstood your comment, you were asking if `simple-storage` had the setItem and getItem, but the documentation doesn't expose any such API.

shared/javascripts/local_storage.js
hitesh96db: :+1:

hitesh96db:
I thought it was not possible to access the `simple-storage` API in privly-applications, but it's possible by making use of XPCOM services as done [here](https://github.com/privly/privly-jetpack/pull/3/files#diff-32500bc6b06ace6093618a1ccae2535bR21) & [here](https://github.com/privly/privly-applications/pull/244/files#diff-d4ff505ec24df0b9ee10bcb964181a00R101).
So that's not a problem :)

hitesh96db:
@smcgregor Should I rebase it ?

smcgregor:
Sure.

I'll be doing more Privly coding/reviewing/merging in the coming week. Can you link me the definitive resources on the different scripting contexts of JetPack and how they fit together? I believe you already linked something for this, but I plan on spending a half day getting up to date and would like to ensure I start at the right place.

hitesh96db:
> API Access for Add-on Code
https://developer.mozilla.org/en-US/Add-ons/SDK/Guides/Two_Types_of_Scripts
> Using XPCOM as a bridge between "chrome" and "sdk"(CommonJS) modules because it's easily available in both environments.
* https://developer.mozilla.org/en-US/docs/Working_with_windows_in_chrome_code#Using_an_XPCOM_singleton_component
* https://developer.mozilla.org/en-US/Add-ons/SDK/Guides/XUL_Migration_Guide#Using_XPCOM
* https://developer.mozilla.org/en-US/Add-ons/SDK/Low-Level_APIs/platform_xpcom

> Complete Guide to Jetpack 
https://developer.mozilla.org/en-US/Add-ons/SDK/Guides

smcgregor:
@hitesh,

Thanks for the resources. I think we can get this merged before I read all of them :)

I think it looks good with the following changes:

* isCommonJS is currently in the global scope and it is uncommented. Is there a reason you need this variable? Can't you drop that if/else block and just use the try/catch since the "catch" block will only be called on JetPack? You can put the `exports.ls = ls;` in the catch block.
* Where you remove items from local storage, i.e. `delete ls.storage[key];`, you should return `undefined`, since that mirrors the built-in local storage.
* I am not certain whether `ss` will be in global scope or not since it is in a catch block. If it is in the global scope, you should rework it.
* I recommend changing `ls.storage` to `ls.jetpackStorage` since it clearly indicates what the variable is referring to in the larger context of storage containers.

Best,
Sean

hitesh96db:
@smcgregor 
>isCommonJS is currently in the global scope and it is uncommented. Is there a reason you need this variable? Can't you drop that if/else block and just use the try/catch since the "catch" block will only be called on JetPack? You can put the exports.ls = ls; in the catch block.

Done.

>Where you remove items from local storage, i.e. delete ls.storage[key];, you should return undefined, since that mirrors the built-in local storage.

was already present :)

>I am not certain whether ss will be in global scope or not since it is in a catch block. If it is in the global scope, you should rework it.

reworked.
>I recommend changing ls.storage to ls.jetpackStorage since it clearly indicates what the variable is >referring to in the larger context of storage containers.

Changed the name to ls.simpleStorage(more accurate).

smcgregor:
Why do you need the two layered try/catch statements? Can't you move the `exports.ls = ls;` into the catch statement and eliminate the nested try/catch?

hitesh96db:
Nope, that doesn't work. I can't use `Components` while exporting it. It's true that the catch block will be called in Jetpack, but while exporting it as a module, it's runtime evironment is that of the sdk, and Components is not defined for the sdk. After importing it, I define simpleStorage as done [here](https://github.com/hitesh96db/privly-jetpack/blob/jpm/lib/local_storage.js#L8).

smcgregor:
I'm confused, why won't this work?

    // Determine whether localstorage can be used directly
    try {
      localStorage;
    } catch(e) {
     
      exports.ls = ls;
     
      // Firefox Jetpack simple-storage 
      ls.localStorageDefined = false;
      const ss = Components.classes["@privly/jetpack;1"].
                  getService(Components.interfaces.nsISupports).
                  wrappedJSObject.
                  getSimpleStorage();
      ls.simpleStorage = ss.storage;
    }

the exports statement is only needed by JetPack and the catch block's exports statement will be called every time on JetPack. No?

hitesh96db:
Yes, the catch block will always be called on Jetpack but there are two cases in which this file is run/used *in Jetpack* -- 
1. I want to reuse this from my add-on scope, which is why I'll load it as a CommonJS module, and hence the presence of exports.
2.This is a normal privileged js file that runs when I open a chrome page that loads this js file, for eg - chrome://login/page. Exports is not defined when run like this, which is why I added the try-catch for that.

hitesh96db:
<pre>
try {
  localStorage;
} catch(e) {
  // Always runs for jetpack
  if (typeof module !== 'undefined' && module.exports) {
    // implies that this is being loaded as a CommonJS module => Case 1
    module.exports.ls = ls;
  } else {
    // 2nd Case.
    ls.localStorageDefined = false;
    const ss = Components...
    ls.simpleStorage = ss.storage;
  }
}
</pre>

hitesh96db:
I'll update the PR, I guess this is more clear.

smcgregor:
Yes, it is more clear, but please add inline comments summarizing the cases that this is covering. "2nd case" references text from this pull request that will not be referenced in the future.

hitesh96db:
@smcgregor added the comments and refactored the code.

smcgregor:
Excellent!

PR #245, opened 2015-05-16
encode privlyLinkKey in Privly link

SummerWish:
according to the code, [privlyLinkKey is decoded when paring](https://github.com/privly/privly-applications/blob/master/shared/javascripts/parameters.js#L47).


smcgregor:
Merged to the development branch. Future pull requests should be made against that branch. https://github.com/privly/privly-applications/tree/develop

PR #241, opened 2015-05-06
Update test to take command line arguments

sammyshj:
Rather than having a separate testing suite for the Safari extension, the Karma Test Runner should support taking command line arguments so that the extension tests can be run using the privly-applications test runner itself.

PR #246, opened 2015-05-17
Add Safari to Unit Tests

smcgregor:
This pull request adds Safari support for TravisCI as well as local tests. A few notes:

* I am merging this with the "develop" branch. Most new pull requests should be made against this branch.
* When launched without specifying a browser the unit tests default to launching Firefox.
* There is a new option for specifying which browsers you want to run the tests on. Before issuing ./run_each.sh you can `export BROWSERS_TO_TEST=Firefox,Chrome,Safari` or any browser subset thereof.
* You can also continue to use Karma's method of specifying browsers when the tests are running locally, `karma start --browsers Firefox,Chrome,Safari`.
* Or you can specify browsers for the SauceLabs script with: `karma start karma.conf-ci.js --single-run --sauce-browsers=Firefox,Chrome,Safari`
* The three different arguments are necessary because `--browsers` will break the SauceLabs script.

Special thanks to @sammyshj for figuring out the argument conflicts.

Closes #241


sammyshj:
:+1:

PR #247, opened 2015-05-17
requires path module to support absolute path for basePath

sammyshj:
Presently, we are trying to run the Safari extension tests using the privly-applications Karma Test Runner. However, when the `karma.conf-ci.js` script is run from a different directory, `fs.existsSync('sauce.json')` searches for `sauce.json` file in that directory. This is fixed in this PR and the `sauce.json` is now searched for in the test directory itself.
When both the karma scripts are run from a different directory, the coverage is reported in the `test/test/coverage/` directory. This PR fixes that and now the coverage is reported in `test/coverage/` directory.

### DIFF
--- a/test/karma.conf-ci.js
+++ b/test/karma.conf-ci.js
@@ -1,4 +1,5 @@
 var fs = require('fs');
+var path = require('path');
 
 module.exports = function(config) {
 
@@ -7,7 +8,7 @@ module.exports = function(config) {
 
   // Use ENV vars on Travis and sauce.json locally to get credentials
   if (!process.env.SAUCE_USERNAME) {
-    if (!fs.existsSync('sauce.json')) {
+    if (!fs.existsSync(path.join(__dirname, 'sauce.json'))) {
       console.log('Create a sauce.json with your credentials based on the sauce-sample.json file.');
       process.exit(1);
     } else {
@@ -30,7 +31,6 @@ module.exports = function(config) {
   // Javascripts.
   var filesToTest = [];
   if (basePath === __dirname) {
-    basePath = '..';
     filesToTest = [
 
       // HTML files to use as a fixtures
@@ -51,6 +51,8 @@ module.exports = function(config) {
       // Test the shared libraries
       'shared/test/*.js'];
   }
+
+  basePath = path.join(basePath, '..');
 
   var filesToExcludeFromTest = [];
   if (process.env.FILES_TO_TEST) {
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -1,5 +1,6 @@
 // Karma configuration
 // Generated on Thu Mar 13 2014 14:12:04 GMT-0700 (PDT)
+var path = require('path');
 
 module.exports = function(config) {
 
@@ -12,7 +13,6 @@ module.exports = function(config) {
   // Javascripts.
   var filesToTest = [];
   if (basePath === __dirname) {
-    basePath = '..';
     filesToTest = [
 
       // HTML files to use as a fixtures
@@ -33,6 +33,8 @@ module.exports = function(config) {
       // Test the shared libraries
       'shared/test/*.js'];
   }
+
+  basePath = path.join(basePath, '..');
 
   var filesToExcludeFromTest = [];
   if (process.env.FILES_TO_TEST) {


smcgregor:
Can you please include descriptions that will help people between 0 and 2 years from now know what this is about? Why do we need support for the absolute path? What is this fixing? etc.

sammyshj:
@smcgregor I have updated the description. After this PR gets merged, I will make necessary changes in the privly-safari repo to update the `run_each.sh` script.

smcgregor:
I think the directory of karma.conf-ci.js needs to be added to NPM's load path because it doesn't find the required libraries when loaded from outside the test directory.

`karma start test/karma.conf-ci.js`

> WARN [reporter]: Can not load "coverage", it is not registered!
>   Perhaps you are missing some plugin?
> WARN [reporter]: Can not load "saucelabs", it is not registered!
>  Perhaps you are missing some plugin?

sammyshj:
@smcgregor This can be solved by setting the `NODE_PATH` environment variable.
We can do something similar to ```export NODE_PATH=`pwd`'/test/node_modules'``` and then run ```karma start test/karma.conf-ci.js``` and it works perfectly fine.

smcgregor:
It would likely be simpler to have the karma script change the cwd to the script's directory. Then nothing else needs to change.

sammyshj:
@smcgregor I don't think I understood the situation here. Can you explain what I am missing?
Currently, when the karma start test/karma.conf-ci.js is run, it looks for the node modules. As there are no node modules present in that directory thats why we have to add the NODE_PATH environment variable.

smcgregor:
The current working directory is almost always on the node path so it is not necessary to change the path if the script is called from the correct place. Since we know the path to the karma script and we have the ability to change the current working directory from that script, it would be simplest to just have the script change the current working directory when it is called.

sammyshj:
@smcgregor I'm trying to point out something else. Initially, the users use `npm install -g karma-cli && npm install`. After that when they run the `karma` command from any other directory where the node_modules is not present, we get an error message shown below:
```
Cannot find local Karma!
  Please install Karma by `npm install karma --save-dev`.
  If you wanna use a global instance, please set NODE_PATH env variable.
```

smcgregor:
The `-g` option should take care of this since it installs Karma globally.

When you issue `karma` in a directory that is unrelated to node and karma, what do you get?

I get this:

screen shot 2015-05-22 at 12 01 32 pm

sammyshj:
After following the same steps the result here is,
screen shot 2015-05-22 at 11 06 20 pm

smcgregor:
It looks like we might need to update the docs? http://stackoverflow.com/questions/20800933/running-karma-after-installation-results-in-karma-is-not-recognized-as-an-inte

sammyshj:
The docs mention that the karma-cli module needs to be installed globally first. So can this PR be merged presently?

smcgregor:
I think it would be better to:

* Update the unit testing docs to include installing the CLI.
* Update the Karma scripts so that it always runs from the testing directory, regardless of where it is called from.

sammyshj:
The docs already include installing the CLI. It can be seen at https://github.com/privly/privly-applications/blob/master/test/README.md#setup

smcgregor:
OK, I'm updating the test runner for privly-chrome at the moment so I can make the script chdir now.

sammyshj:
@smcgregor Okay. Does similar changes need to be made in the privly-safari repo too?

smcgregor:
So `process.chdir(__dirname);` doesn't include the __dirname in the NODE_PATH. I am going to make more substantive changes to the testing script so it doesn't assume any files will be loaded.

I think the Safari changes will be obvious once I open a pull request.

sammyshj:
Can we do something similar to `process.env.NODE_PATH = __dirname;`?

smcgregor:
That didn't work when I tested it. I think NODE_PATH is read in before executing the script.

sammyshj:
@smcgregor I think we somehow need to export NODE_PATH as it is being checked in the karma-cli script. The script can be seen here https://github.com/karma-runner/karma-cli/blob/master/bin/karma#L8
As the NODE_PATH environment variable is not set, I'm seeing the error at https://github.com/karma-runner/karma-cli/blob/master/bin/karma#L67

smcgregor:
Right now I am issuing it with a command like: `NODE_PATH=/Users/smcgregor/Projects/privly-chrome/privly-applications/test/node_modules ./tests/run_each.sh`

but the addition of the NODE_PATH should be unnecessary.

I haven't tested syntax like the one you linked, so I'll give that a try.

smcgregor:
I integrated the test runner with Chrome [here](https://github.com/privly/privly-chrome/pull/123), which requires the changes for privly-applications [here](https://github.com/privly/privly-applications/pull/248).

I think the changes in the privly-applications pull request should take care of your issues, so long as you reference scripts and test scripts relative to privly-applications/. Does that sound good?

sammyshj:
@smcgregor I have tested both the PRs on my local machine and the tests pass successfully. I have mentioned the same on the PRs too and have given a :+1: on both of them.
Once these PRs get merged, I will update the privly-safari repo too and make similar changes as that of privly-chrome repo to make use of one single Karma Test Runner

PR #248, opened 2015-05-22
Update chrome testing

smcgregor:
This pull request updates the karma testing scripts to make it easier for them to be used from different extension contexts by removing the shared libraries from the karma scripts into the run_each.sh scripts.

When testing from an extension, you will likely need to export the node path to the testing modules:

`export NODE_PATH=/PATH/TO/privly-applications/test/node_modules`


sammyshj:
I have tested this on my local machine and it seems to be working fine :+1:

PR #250, opened 2015-05-25
Refactor options and glyph icons

SummerWish:
See https://github.com/privly/privly-chrome/pull/124 for details.

It would be great if you merge PR timely because there are some code conflicts between this PR and my embed-posting functionality. I may forget some details and make mistakes if I process those duplicates after a long time :-)


SummerWish:
Currently `options.js` and `glyph.js` is not highly unit test covered. I will add more tests gradually.

shared/javascripts/options.js
```diff
@@ -0,0 +1,387 @@
...
+    optionChanged('options/injection', enabled);
+    return true;
+  };
+
+  /**
+   * Validates a FQDN
+   * @param  {String}  domain Domain to validate
+   * @return {Boolean}
+   */
+  Privly.Options.isWhitelistDomainValid = function (domain) {
```
smcgregor: This isn't checking that a domain is whitelisted, it is checking that the domain is valid. The function's name should be "isValidDomain".

Help/test/new.js
```diff
@@ -39,7 +39,7 @@ describe ("Help New Suite", function() {
     // Call the test
     callbacks.pendingLogin();
 
-    expect(ls.getItem("Login:redirect_to_app")).toBe(window.location.href);
+    expect(Privly.storage.get("Login:redirect_to_app")).toBe(window.location.href);
```
smcgregor: The difference in conventions between "Login:redirect_to_app" and "Login/redirect_to_app" bothers me. I think we need to adopt one or the other. I am OK with moving to the "/" convention.

smcgregor:
[Updated coverage :)](https://coveralls.io/builds/2689611)

SummerWish:
LOL That's great!

Help/test/new.js
SummerWish: +1

smcgregor:
@hitesh96db @irdan @sammyshj I intend on merging this shortly. It will break several features on Firefox Xul and the developing Jetpack version related to setting and getting options in various contexts. 

@hitesh96db should continue using the gsoc-jetpack branch for development. It should not be difficult to merge it into the development branch later since it will only require adding the JetPack message API.

@sammyshj I believe you haven't implemented all the options functionality yet, but when you implement setting options you should look at the points where `chrome.runtime.onMessage.addListener` and `chrome.runtime.sendMessage` are called in this pull request since the Safari extension will need to mirror these with Safari's message API.

I think we need a unified message passing library in shared/javascripts/ for platform agnostic JSON message passing. Any volunteers? I could also take care of this fairly quickly.

smcgregor:
[More documentation on message passing.](http://privly.github.io/pages/develop.html#ScriptingContextMessages)

smcgregor:
All,

I opened pull requests against @SummerWish's pull requests for refactoring how we send messages between different contexts since the old method is not necessary once we are off Xul and it was getting unmaintainable. I highlighted places where people need to make changes with console messages for Jetpack and Safari. This should make it _much_ simpler to integrate these two platforms since it mostly comes down to data munging.

I invite review at this point since I have not had time to look at this with fresh eyes and I still need to work on testing. Regardless, I am eager to land this in develop ASAP since it is a dependency for 3 different GSoC projects.

https://github.com/SummerWish/privly-chrome/pull/1
https://github.com/SummerWish/privly-applications/pull/1

Best,
Sean

smcgregor:
Should I take a look at the current version? Do you have a diagram from the changes you made?

SummerWish:
Hi, I refactored the message part according to this diagram (names are not exactly the same):

diagram

smcgregor:
Have you run the selenium tests on this yet? I think the next step will be to write and run integration tests that use each of these message pathways. Unit tests won't test this very well.

SummerWish:
Sorry not yet.. I'm afraid I don't have time to write integration tests these days, I will add them later. Working hard on writing a 8000 word paper (homework) these days.. 😭

smcgregor:
@hitesh96db and @sammyshj Can you take a look at this pull request and let me know if you foresee any issues integrating it with Jetpack/Safari? Once I merge it you will be responsible for integrating it with your respective platforms.

hitesh96db:
@smcgregor I'm at San Francisco for Google Code In trip, so I won't be able to look at it for another 4 days. If this needs to be merged, you could do that and I'll integrate it with jetpack once I reach home.

smcgregor:
Sounds good. How long are you there? I might be grabbing breakfast in the Mountainview area on Friday since I'm driving across the country.

hitesh96db:
I'm at MTV right now, but I'll be leaving on Thursday.

smcgregor:
Cool, have fun :)

sammyshj:
@smcgregor Yes, I will pull this branch onto the privly-safari repo and let you know if I face any problems with integrating the sendMessage and onMessage.addEventListener.
I will be opening a new PR on the privly-safari repo with basic popover implementation shortly.

sammyshj:
@SummerWish I have just made added the new files from privly-applications to my fork of privly-safari https://github.com/sammyshj/privly-safari/commit/44f878a867941f95c84960129adb9a8f87b252e5
On checking the browser console, I get the warning from https://github.com/SummerWish/privly-applications/blob/ed79031cf54910eca48fc2a03dce04e82a261ef7/shared/javascripts/context_messenger.js#L198
I still have not implemented the popover in the Safari extension. Can you let me know if this is normal?

SummerWish:
@sammyshj Hi, that's OK. It is because for your current platform, no suitable adapter is found. So the context messenger fallbacks to the `HostedAdapter` and you will get that warning when `setListener()` is called.

SummerWish:
@sammyshj You may need to change `isPlatformMatch` here: https://github.com/SummerWish/privly-applications/blob/ed79031cf54910eca48fc2a03dce04e82a261ef7/shared/javascripts/context_messenger.js#L324 . write platform detection code in it. Then after you implemented `setListener` you will see your `setListener` being called.

sammyshj:
@SummerWish Thanks for pointing that out. I have added a basic commit https://github.com/sammyshj/privly-applications/commit/ea8f863a31d2cc0a9722a7636762300613211a97 to add the safariAdapter setListener function. On running the privly safari extension, the expected behaviour can be seen.
Next, I will actually have the setListener function do something to make sure that the other functions in context_messenger.js are working fine for Safari.

smcgregor:
I'll likely merge this in the next day so @SummerWish can base his changes on this. It still needs more tests before I will be comfortable merging it to master.

SummerWish:
Hi all @smcgregor @sammyshj @hitesh96db, I just pushed some commits, which might be (almost) the last interface change of `Privly.message`:

1. `Privly.message.messageTopPrivlyApplications` changed to `Privly.message.messagePrivlyApplication`.
   Messages can be sent to non-top privly applications now.

2. `Privly.message.messageExtensions`, `Privly.message.messageContentScripts`, `Privly.message.messagePrivlyApplications` are using `Promise` now. All responses are returned as promise resolve value. This can simplify the pattern of `message and receive response`.
   Notice that even when you don't explicitly send a response when receiving messages, you will get a resolved Promise object after sending the message. This can make your promise handling code more consistent.

   Example:

   ```js
   Privly.message.messageExtension('some data').then(function () {
       console.log('message sent!');
   });

   Privly.message.messageExtension('some data which will receive response').then(function (data) {
       console.log('message sent and got response:', data);
   });
   ```

3. The behavior of `Privly.message.addListener` is now [similar to Chrome.runtime.onMessage.addListener](https://developer.chrome.com/extensions/runtime#event-onMessage). In detail, your listener function accepts two parameters, `data` and `sendResponse`. You can use `sendResponse` function to send a response back to the sender. If you want to send a response asynchronously, you need to return `true` in your message handler.
   Notice that, the original behavior of returning `true` is to remove the message handler. This can be accomplished by `Privly.message.removeListener` now.

   Example:

   ```js
   Privly.message.addListener(function (data, sendResponse) {
       setTimeout(function () {
           sendResponse('response');
       }, 100);
       return true;  // if you want to send async response, return true here
   });

   Privly.message.addListener(function (data, sendResponse) {
       return true;  // dont do this. this will leave the message channel open and cause memory leak
   });
   ```

SummerWish:
I just pushed some integration test file for context messenger. Could you have a look?

hitesh96db:
Hey @smcgregor @SummerWish , I just went through this PR and realized that the context messenger needs to be refactored/redesigned because the Jetpack messaging architecture is different from Chrome. 
The differences are -
* In Chrome, I can listen to a general message event(which is what I assume `chrome.runtime.onMessage.addListener` does). In Jetpack, I have to set listeners to a specific message name/identifier, for eg - `self.port.on("foobar", callback)` listens for messages sent with name "foobar".
* In Chrome, the API object that is used to listen to or send messages is known and done via `chrome`. In Jetpack, the object used to send or listen to messages changes depending on the context and therefore there will be a need to pass the object as a parameter in the functions that listen to or send messages.

Mostly, the function parameters might change. I think the adapters should play a part in `Privly.message.(remove/add)Listeners` as well and override it's behaviour if necessary. I'll try to come up with something more concrete.

SummerWish:
Could we use a hard coded name for issue 1? Like "__Privly__"?

For issue 2, could you just write if statements for different context? Your send message interface could call getContext() to know the current context.

hitesh96db:
Ah, I might not have framed the issue 2 properly. Let's say, I create a popup in Jetpack and the object created is saved as `popupButton`. This object(`popupButton`) allows me to access the properties of the popup such as height, width, etc and allows me to send messages to the content scripts injected in popup.html. This object is what I need to communicate with content scripts. Similarly, there are other objects such as [page-mod](https://developer.mozilla.org/en-US/Add-ons/SDK/High-Level_APIs/page-mod#Communicating_With_Content_Scripts). Even in a specific context, there are several objects which are not known beforehand.

Yeah, I think a hard coded message name might work.

SummerWish:
Not familiar with Jetpack, is it possible to store those in a global scope and access them in your adapter? Anyway, changes to the context messenger are welcomed 😄

sammyshj:
@SummerWish I have implemented the other functions for the Safari browser. This can be seen at https://github.com/sammyshj/privly-applications/commit/c42907edacc7aa5eadbb6931a567b39189bcefd7
The `sendMessageTo` function has been implemented so now, the messages can be sent. I have tested this by adding `addEventListener ` to the files.
However, I am not completely sure what the `setListener` function should contain. Should the scripts have a `Privly.message.addListener` for receiving the messages?
In Safari, for receiving messages in different context, different commands are used.
For receiving messages from an Injected Script, `safari.application.addEventListener` should be used whereas for receiving messages in an Injected Script, `safari.self.addEventListener` should be used.
More about this can be read at https://developer.apple.com/library/safari/documentation/Tools/Conceptual/SafariExtensionGuide/MessagesandProxies/MessagesandProxies.html
As we are referring to `Privly` variable from the background html page, does this mean that only the `safari.self.addEventListener` should be used?

SummerWish:
Scripts call `Privly.message.messageXXX` to send messages and `Privly.message.addListener` to listen messages. In your adapter, you should implement `setListener`.

how about such code:
```js
SafariAdapter.prototype.setListener = function (callback) {
    if (this.getContextName() === 'CONTENT_SCRIPT') {
        safari.self.addEventListener(.....);
    } else {
        safari.application.addEventListener(......);
    }
}
```

hitesh96db:
@SummerWish @sammyshj @smcgregor 
Is it Ok to add names to messages and listen to those specific messages only ?. Instead of calling all the functions added as a listener [here](https://github.com/privly/privly-applications/pull/250/files#diff-e3c7938b324cb640951cd58dd4d34be2R674), only that function will be called for which the message is intended. Will that work for Chrome or Safari ? Often only one function among the listeners would need to process the message(because the message is intended only for that context/function). We can add the name provided by the user in the header, and later process it.

Also, adding listeners like `Privly.message.addListener(function(){ console.log("Hello!"); })` adds duplicates. Saving the function in an object and then calling addListener would work but is there a way to avoid duplication in such a case ?

smcgregor:
@hitesh96db To be cross-platform the name of the desired function would need to be in the payload (JSON document) of the message. We could potentially establish an explicit convention where the sender always specifies the expected handler in the receiving context and only one handler is ever called, but I think it is better for the handler itself to make this determination.

I can't think of an instance in which duplicate listeners may be added where the problem wouldn't be solved by cleaner code.

@SummerWish @hitesh96db @sammyshj I think this is getting sufficiently drawn out that I am going to merge this into a branch, "gsoc-2015," that will be all of your privly-applications development branches. It feels like there is going to be too much churn to push to master or develop at this point.

hitesh96db:
@smcgregor Any reason for leaving it to the handler to determine if message is the intended one or not ? I think it'll be better for the messaging interface to take care of this, the adapters do more of the work then, making it easier to write the handlers.

hitesh96db:
(Jetpack)
Also , now with `context_messenger.js`, that would mean one more script injected along with the `privly.js` on every small/big page load event. I suppose that's not a good thing.

smcgregor:
There are two potential models here. Either messages are "events," which are broadcast to the scripting context and each handler can decide whether it wants to act on the data, or the events are cross-environment function calls where the function knows it is the intended recipient of the message. A well-written handler will still check the message to see if it is the intended recipient and it has the data it needs.

I am indifferent between the two philosophies, except we currently have an implementation without named handlers. I don't think making the handler name explicit will make it easier for the developer because the handler should be written the same regardless of whether it was named in the message... but if this is causing you some implementation burden, please respond back with specifics so I can understand the issue you are having.

hitesh96db:
No implementation burden. I can work with both, I just thought as to why named handlers wasn't considered.
Anyways, there are still some changes to be made for this to work with Jetpack. I'll send a PR soon but you could merge this one if you're planning to.

sammyshj:
@SummerWish Can you add commits https://github.com/sammyshj/privly-applications/commit/c42907edacc7aa5eadbb6931a567b39189bcefd7 and https://github.com/privly/privly-applications/commit/82d5068ece298a1aad27383953641ef8a5738edf to the same PR?

SummerWish:
@sammyshj merged

SummerWish:
@sammyshj You can create send PR to the `options-interface` branch on https://github.com/SummerWish/privly-applications/  :-)

sammyshj:
@SummerWish Thanks for that. I have tested the code from the BACKGROUND_SCRIPT. I am currently looking for the SafariAdapter to work completely with the CONTENT_SCRIPT too and if there are any changes, I will open a new PR as you mentioned. Thanks

hitesh96db:
I've made some changes to the interface, [PR](https://github.com/SummerWish/privly-applications/pull/2/).

SummerWish:
Hi all @sammyshj @hitesh96db @smcgregor 
Due to limitations of broadcasting messages and receiving a response, now you have to explicitly pass a parameter if you wish to receive a response after sending(broadcasting) a message (`messageXXX()`).

This is just a reminder of a small interface change. Code of your platform adapters needn't change anything :-)

smcgregor:
I am merging this to a branch called `gsoc-develop`, which I hope to merge to develop fairly soon.

There are currently many errors: https://travis-ci.org/smcgregor/privly-applications/builds/67769292

I'll start working through the errors today, but feel free to start examining them.

smcgregor:
Running an updated build [here](https://travis-ci.org/smcgregor/privly-applications/builds/67777368).

smcgregor:
@SummerWish It looks like we will need to use the polyfill for promises after all because the market share is poor enough that the hosted apps will not work for many people. Please don't use newer standards in the future -- I don't trust polyfills and will be retiring the polyfill ASAP.

smcgregor:
@sammyshj I [increased Safari's version number](https://github.com/smcgregor/privly-applications/commit/3e18644296920161e6aba403c6909b4b4c6bb8ef) because the prior version didn't define promises.  Now, [SauceLabs](https://saucelabs.com/tests/3b83f0f50a02456a95004f90395bb1d4) seems [to disconnect](https://travis-ci.org/smcgregor/privly-applications/builds/67780768). Any ideas?

sammyshj:
@smcgregor According to https://docs.saucelabs.com/reference/platforms-configurator/#/ the current configuration should be working fine. I will run the tests from my machine on SauceLabs and see what the output is.

sammyshj:
@smcgregor I have tried running the tests(the present tests on the develop branch) on Safari 8.0 on OS X 10.10 and they seem to run successfully. The test session was https://saucelabs.com/tests/fcac0c4157514143ae20cda7e8977263
screen shot 2015-06-22 at 6 36 22 am

smcgregor:
@sammyshj I removed most of the tests and ran it on SauceLabs for safari from my local machine. After repeated attempts it seems to fail after a random number of tests. Is this the case for you as well when you run my gsoc-develop branch?

sammyshj:
@smcgregor Yes, it is the same case for me too.
When I run the tests on my local Safari browser then, all the tests run successfully. However, on running on SauceLabs, the tests get 'DISCONNECTED' at times.
It looks like there is some issue with SauceLabs in this matter?

SummerWish:
Thank you. BTW is it ok to only add polyfill for safari? Because Chrome and Firefox has a good supporting for Promise.

smcgregor:
And IE. We don't have an extension for it, but people will read the content on the hosted fallback.

SummerWish:
Oh. I see.

smcgregor:
@sammyshj Which version of Safari do you run locally? I don't have 8.0 on my machine so I can't tell whether this is SauceLabs or the version of Safari. Once we have the polyfill in place I'll downgrade the Safari version for SauceLabs and wait to see if SauceLabs fixes this.

sammyshj:
@smcgregor The version of Safari running locally is, `Version 8.0.4 (10600.4.10.7)` and the tests run good locally.

smcgregor:
@SummerWish, @hitesh96db, and @sammyshj 

I pushed my [gsoc-develop branch](https://github.com/privly/privly-applications/tree/gsoc-develop) to privly/privly-applications. You should base your work on this branch and make pull requests against it. Once the transition to the context messenger is complete across all relevant platforms, I will open a pull request to merge into develop.

Notes:

* all tests currently pass, but the Safari tests will fail on SauceLabs because it loses the connection.
* I merged in all commits from this PR and from the develop branch.

-Sean

History/js/new.js
```diff
@@ -5,24 +5,6 @@
...
-  /**
-   * Attach the message listeners to the interface between the extension
-   * and the injectable application.
-   */
-  initialize: function() {
-    privlyExtension.initialContent = function(){};
-    privlyExtension.messageSecret = function(){};
-      
-    // Initialize message pathway to the extension.
-    privlyExtension.firePrivlyMessageSecretEvent();
```
hitesh96db: @SummerWish No more message secret ? I think it might be needed for FF since messages passed b/w content scripts and privly applications will happen via `parent.postMessage`

smcgregor: I believe I removed this. It is better we assume the content script is controlled by the host page and only use the content script <-> privly app message pathway for messages that are not privileged, like content height.

PR #257, opened 2015-05-28
Added safari_web for selenium tests

sammyshj:
Adds safari_web for selenium tests. This way we can run the selenium tests that already run on Chrome and Firefox to run on Safari too.

### DIFF
--- a/test/selenium/configurations.rb
+++ b/test/selenium/configurations.rb
@@ -127,6 +127,13 @@ def configure_for_chrome_web(args)
   common_configuration_for_web(args)
 end
 
+def configure_for_safari_web(args)
+  common_configuration_for_web(args)
+  # Do not block pop-up windows in Safari
+  # https://macmule.com/2012/07/31/disabling-safari-5-1-xs-6-xs-pop-up-blocker-from-terminal-2/
+  `defaults write com.apple.Safari com.apple.Safari.ContentPageGroupIdentifier.WebKit2JavaScriptCanOpenWindowsAutomatically -bool true`
+end
+
 def configure_for_sauce_firefox_web(args)
   common_configuration_for_sauce
   common_configuration_for_sauce_web(args)
--- a/test/selenium/specs/tc_new.rb
+++ b/test/selenium/specs/tc_new.rb
@@ -10,6 +10,7 @@ class TestNew < Test::Unit::TestCase
   def setup
     page.driver.browser.get(@@privly_test_set[0][:url])
     login(@@privly_test_set[0][:content_server])
+    page.driver.browser.manage.timeouts.implicit_wait = 3
   end
 
   # Test post creation for each application that presents a "new" action.
@@ -66,7 +67,7 @@ class TestNew < Test::Unit::TestCase
         find_button("Update").click
 
         # Refresh
-        page.driver.browser.get(page.evaluate_script("window.location.href"))
+        page.driver.browser.navigate.refresh
 
         # Make sure the refreshed page has the new content
         assert page.has_text?('Updated!')
--- a/test/selenium/run_all.rb
+++ b/test/selenium/run_all.rb
@@ -17,7 +17,7 @@ args = {}
 optsHelp = ""
 OptionParser.new do |opts|
   opts.banner = "Usage: ruby run_all.rb [options]"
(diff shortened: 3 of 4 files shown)


smcgregor:
When I watch selenium interact with the browser I see that the link was clicked, but a new window is not opening. I don't have more time to invest in this at the moment, so let's merge this knowing the test fails and we will fix it when we integrate SauceLabs.

sammyshj:
@smcgregor Sounds good. I am currently working on getting selenium to work with the Safari browser locally.

sammyshj:
@smcgregor I have modified the code to add an implicit_wait to avoid the Element not found error on my local machine. Also, I have removed the Timeout.timeout and added wait instead as this does the desired and is commonly used in selenium.

sammyshj:
@smcgregor The selenium tests were failing on Safari and now pass successfully.
The reason why the tests were failing is that the `Block pop-up windows` option was checked in the Safari browser. We can uncheck that option from, Preferences -> Security.
We can also do the same from the terminal which has been added in my last commit https://github.com/sammyshj/privly-applications/commit/63e24d03f8aa09e598c215f340461991485ee080
If the changes look good, I will squash the commits for the PR to be merged

test/selenium/specs/tc_new.rb
```diff
@@ -10,6 +10,7 @@ class TestNew < Test::Unit::TestCase
   def setup
     page.driver.browser.get(@@privly_test_set[0][:url])
     login(@@privly_test_set[0][:content_server])
+    page.driver.browser.manage.timeouts.implicit_wait = 3
```
smcgregor: Do you still need this change if you change the safari option in the terminal?

test/selenium/configurations.rb
```diff
@@ -127,6 +127,12 @@ def configure_for_chrome_web(args)
   common_configuration_for_web(args)
 end
 
+def configure_for_safari_web(args)
+  common_configuration_for_web(args)
+  # Do not block pop-up windows
+  `defaults write com.apple.Safari com.apple.Safari.ContentPageGroupIdentifier.WebKit2JavaScriptCanOpenWindowsAutomatically -bool true`
```
smcgregor: This somehow changes the configuration of Selenium? Do you have documentation for this?

test/selenium/specs/tc_new.rb
sammyshj: It is better to have an implicit wait to avoid NoSuchElementError but, the tests pass successfully even without having the implicit_wait.
Should this line be removed then?

test/selenium/configurations.rb
sammyshj: https://github.com/sammyshj/privly-applications/blob/updateselenium/test/selenium/configurations.rb#L133 enables pop-up windows in Safari.
This can be done by opening Safari and navigating to, Preferences -> Security -> Uncheck block pop-up windows.
But I thought that rather than the user doing the whole thing, we can simply run this from the script itself.
Do you think this line should be removed or should be put into some other file?

test/selenium/specs/tc_new.rb
sammyshj: On my local machine, the tests sometimes fail when the implicit_wait is not mentioned.

test/selenium/configurations.rb
smcgregor: I don't understand how this command is changing the safari setting. It is fine as-is, but I want to know how it works.

test/selenium/specs/tc_new.rb
smcgregor: I'm wondering whether Capybara might not like this, but if it passes on Safari this should be fine.

test/selenium/configurations.rb
sammyshj: The defaults command in OSX let users access the Mac OS X user defaults system. This can be seen at https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/defaults.1.html
More about the terminal command can be seen at https://macmule.com/2012/07/31/disabling-safari-5-1-xs-6-xs-pop-up-blocker-from-terminal-2/

smcgregor: OK, once you put this information in an inline comment, I'll merge.

sammyshj: Added the link in the comment and squash the commits into one.

PR #259, opened 2015-05-31
Back to prefs

hitesh96db:
`simple-storage` didn't work :(
The XPCOM component that allows us to use simple-storage in chrome code isn't recognized when chrome URLs are embedded in iframes.
So back to `prefs` for the time being before shifting to `indexedDB`.


hitesh96db:
In short, 
`simple-storage` didn't work.
The XPCOM component that allows us to use simple-storage in chrome code isn't recognized when chrome URLs are embedded in iframes.
So back to `prefs` for the time being before shifting to `indexedDB`.

PR #260, opened 2015-06-02
Added basic configuration for safari_extension

sammyshj:
Added configuration for safari_extension. This helps in running the selenium tests on Safari locally with the extension.

### DIFF
--- a/shared/javascripts/network_service.js
+++ b/shared/javascripts/network_service.js
@@ -99,6 +99,8 @@ var privlyNetworkService = {
       return "CHROME";
     } else if(window.location.href.indexOf("chrome://") === 0) {
       return "FIREFOX";
+    } else if(window.location.href.indexOf("safari-extension://") === 0) {
+      return "SAFARI";
     } else {
       return "HOSTED";
     }
@@ -247,6 +249,7 @@ var privlyNetworkService = {
       return protocolDomainPort;
     } else if (platformName === "CHROME" ||
                platformName === "FIREFOX" ||
+               platformName === "SAFARI" ||
                platformName === "IOS") {
       return ls.getItem("posting_content_server_url");
     } else if (platformName === "ANDROID") {
--- a/test/selenium/README.md
+++ b/test/selenium/README.md
@@ -85,6 +85,9 @@ this from the privly-applications directory of the privly-firefox extension.
 **chrome_extension:** run Chrome locally with an extension. You should only run
 this from the privly-applications directory of the privly-chrome extension.
 
+**safari_extension:** run Safari locally with an extension. You should only run
+this from the privly-applications directory of the privly-safari extension.
+
 You can run each of these from the `privly-applications/test/selenium` directory with
 `ruby run_all.rb -p INSERT_PLATFORM_HERE -r experimental -c http://localhost:3000`.
 The corresponding browser will be opened and the specs included by `run_all.rb` will be run.
@@ -137,7 +140,7 @@ From the privly-applications directory, run: `ruby run_all.rb -p sauce_chrome_ex
 This will launch Safari on Windows. It will navigate the
 webserver you have running on localhost so make sure that is where the applicaion files are that you want to test.
 
-From the privly-applications directory, run: `ruby run_all.rb -p sauce_chrome_web -c http://localhost:3000 -r experimental`
+From the privly-applications directory, run: `ruby run_all.rb -p sauce_safari_web -c http://localhost:3000 -r experimental`
 
 ## TravisCI
 
--- a/test/selenium/configurations.rb
+++ b/test/selenium/configurations.rb
@@ -227,3 +227,12 @@ def configure_for_chrome_extension(args)
   assign_chrome_extension_path
 
 end
+
+def configure_for_safari_extension(args)
+  Capybara.register_driver :safari_extension do |app|
+    Capybara::Selenium::Driver.new(app, :browser => :safari)
+  end
+  @@privly_applications_folder_path = ""
+  Capybara.current_driver = :safari_extension
+  Capybara.default_driver = :safari_extension
+end
--- a/test/selenium/configurations.rb
+++ b/test/selenium/configurations.rb
@@ -55,6 +55,16 @@ def assign_chrome_extension_path
   @@privly_applications_folder_path = ""
 end
 
+def assign_safari_extension_path
+
+  # The safari extension URL changes everytime Safari is opened so,
+  # it is grabbed from the privly test page's injected application.
+  # It also fixes the path in the CRUD tests since this test doesn't
+  # run until all the initialization is complete
+  require_relative "tc_safari_helper"
+  @@privly_applications_folder_path = ""
+end
+
 # This is the common config for running tests from the
 # web scripting environment on SauceLabs
 def common_configuration_for_sauce_web(args)
@@ -232,7 +242,9 @@ def configure_for_safari_extension(args)
   Capybara.register_driver :safari_extension do |app|
     Capybara::Selenium::Driver.new(app, :browser => :safari)
   end
-  @@privly_applications_folder_path = ""
   Capybara.current_driver = :safari_extension
   Capybara.default_driver = :safari_extension
+
+  assign_safari_extension_path
+
 end
--- a/test/selenium/run_all.rb
+++ b/test/selenium/run_all.rb
@@ -17,7 +17,7 @@ args = {}
 optsHelp = ""
 OptionParser.new do |opts|
   opts.banner = "Usage: ruby run_all.rb [options]"
-  opts.on('-p', '--platform PLATFORM', 'The target platform (firefox_web, firefox_extension, chrome_web, chrome_extension, safari_web, sauce_chrome_web, sauce_firefox_web, sauce_safari_web, sauce_firefox_extension, sauce_chrome_extension)') { |v| args[:platform] = v }
+  opts.on('-p', '--platform PLATFORM', 'The target platform (firefox_web, firefox_extension, chrome_web, chrome_extension, safari_web, safari_extension, sauce_chrome_web, sauce_firefox_web, sauce_safari_web, sauce_firefox_extension, sauce_chrome_extension)') { |v| args[:platform] = v }
   opts.on('-r', '--release-status RELEASE', 'The target release stage (experimental, deprecated, alpha, beta, release)') { |v| args[:release_status] = v }
   opts.on('-c', '--content-server SERVER', 'The content server (http://localhost:3000)') { |v| args[:content_server] = v }
   optsHelp = opts.help
@@ -106,6 +106,7 @@ end
 #                    This requires a webserver to run on localhost:3000
 #  safari_web: run Safari locally without an extension.
 #              This requires a webserver to run on localhost:3000
+#  safari_extension: run Safari locally with an extension.
 if args[:platform]
   platform = args[:platform]
 end
@@ -170,6 +171,9 @@ elsif platform == "chrome_extension"
 elsif platform == "sauce_chrome_extension"
   sanity_check "privly-chrome"
   configure_for_sauce_chrome_extension(args)
+elsif platform == "safari_extension"
+  sanity_check "privly-safari"
+  configure_for_safari_extension(args)
 else
   puts "The platform you selected was not recognized or is not supported"
   puts "Note that Safari and Internet Explorer are integrated"
--- a/test/selenium/run_all.rb
+++ b/test/selenium/run_all.rb
@@ -183,4 +183,8 @@ end
 initialize_CRUD_tests(args)
 
 # Run all the tests
-require_relative "specs/ts_all_specs"
+if platform == "safari_extension"
+  require_relative "specs/tc_safari"
+else
+  require_relative "specs/ts_all_specs"
+end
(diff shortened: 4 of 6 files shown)


sammyshj:
@smcgregor We can now get the extension's base URL but, on running the tests, I get an error saying, 'content server is unavailable'. Can you test it on your local machine and let me know if this is the case for you too?

smcgregor:
Are you running a content server at the address you pass for the `-c http://localhost:3000` parameter?

sammyshj:
Yes, I am running the content server and it can be seen at http://localhost:3000

smcgregor:
Are you seeing any requests in the content server's console when you run the tests?

sammyshj:
No, I don't see any requests in the content server's console.

smcgregor:
Next set the Capybara timeout to 60 seconds so you have enough time to look at the error console while selenium is controlling the browser. See if there is an error when it should be asking the server to login.

smcgregor:
https://github.com/privly/privly-applications/blob/master/test/selenium/configurations.rb#L13

sammyshj:
@smcgregor This is where the tests get stuck.
screen shot 2015-06-05 at 10 57 52 pm

smcgregor:
Yes, but what does the browser console say?

sammyshj:
The console output can be seen at https://gist.github.com/sammyshj/d5a415b5da0d2058f432

smcgregor:
That is the console, I am asking about the browser console. I think you can get there by:

1. Open Safari and choose Safari > Preferences > Advanced. Then select the checkbox 'Show Develop menu' in menu bar.
2. Choose Develop > Show Error Console.

sammyshj:
The output of the error console is,
screen shot 2015-06-05 at 11 31 14 pm
The request for the user_account_data does not have a complete extension base URL.

smcgregor:
It looks like you set the content server's URL to be the address of the extension? The URL should be http://localhost:3000/posts/user_account_data, not com.privly.privly-...

sammyshj:
@smcgregor I'm currently looking at https://github.com/sammyshj/privly-applications/blob/updateselenium/test/selenium/tc_safari_helper.rb#L20 if this is where things should be changed.
Also, the error can be because of this, https://github.com/privly/privly-applications/blob/master/shared/javascripts/network_service.js#L117

smcgregor:
I think the problem is that you are running tests before you set the content server the extension is associated with. In many of the tests, [this helper](https://github.com/privly/privly-applications/blob/master/test/selenium/specs/auth_helper.rb#L10) is used to initialize the content server before doing any testing.

sammyshj:
@smcgregor Can you replicate the same error on your local machine?

smcgregor:
Probably, I don't need to. I think you can solve it by implementing [this](https://github.com/privly/privly-applications/blob/master/shared/javascripts/network_service.js#L240) for Safari.

smcgregor:
(I am in and out of meetings most of the day)

The problem is your privly applications think the content server they should be talking to is the same as the URL they are being viewed at. This is not the case when the application is served by the extension. If you hard-code the address [here](https://github.com/privly/privly-applications/blob/master/shared/javascripts/network_service.js#L240), you should be able to test it.

sammyshj:
@smcgregor Yes, Im working at it right now.
Thanks

sammyshj:
@smcgregor I have modified the code with commit, https://github.com/sammyshj/privly-applications/commit/76f276564b5348c16fb31464f767a4d2b622e199
Now, the previous error of bad URL in the browser console does not appear anymore. However, the tests which require login still don't pass.

smcgregor:
What does the server's console say?

sammyshj:
@smcgregor I have added the SAFARI platformName in https://github.com/sammyshj/privly-applications/commit/e6179a7864e21a96a11e0ef2ba40d2c173adfc04

sammyshj:
@smcgregor The test page is now redirected to privlyalpha.org Help page and gets stuck there.

sammyshj:
@smcgregor The tc_show_injected tests run successfully, but, the tests which have to navigate to the extension URL fail.

smcgregor:
Make sure you explicitly use the selenium driver for navigation tasks since
it fails on other extensions when using the capybara statement. Something
like page.driver... (I'm on my phone at the moment)

sammyshj:
Yes, for navigation, the command is page.driver.browser.get()
This is already being used in the specs directory tests.
I have found a similar error issue, https://github.com/angular/protractor/issues/2111
The statements after page.driver.browser.get() are not executed. However, when I replace page.driver.browser.get() with window.location.replace(), the next statements are then executed.

smcgregor:
Conditionally replace the driver.get statement with window.location when
the platform is safari_extension.

sammyshj:
@smcgregor The tests still dont seem to pass. However the same test passes and the user is logged in if the argument passed to the test is safari_web.
Please let me know if the same is the case in your local machine too.

sammyshj:
@smcgregor The reason why the authentication script does not work is, because the page still seems to be on http://test.privly.org/test_pages/url_parameters.html even when it reaches the auth_helper.rb script.

sammyshj:
@smcgregor The script seems to now have the correct extension URL and the correct content server URL but, even when the auth_helper.rb script is run, the browser is still stuck at URL http://test.privly.org/test_pages/url_parameters.html
I have tried modifying the code to add explicit wait but, the browser still seems to be stuck at the URL.
Can you let me know if you are facing the same situation?

smcgregor:
I ran the selenium tests locally and some of them are passing. `window.location.replace` gives an error when I try to use it.

To select which tests run, you should comment out lines in the `ts_all_specs.rb` file. This will make it easier to debug individual test cases. You should not expect all the tests to pass initially since these tests have never been run on Safari. In some cases the tests may fail because there is a problem with privly-applications or privly-safari, but if my experience is any indication of the future you can expect the tests to fail just because you are integrating several systems that have not been integrated before and the syntax for Selenium and Capybara will be slightly different.

When I navigate directly to a Safari extension URL using `page.driver.browser.get` it looks like Selenium may be losing connectivity with the browser, as evidenced by `Selenium::WebDriver::Error::TimeOutError: timed out waiting for Safari to respond`. My working theory is that Safari rejects the Selenium WebDriver browser extension when the scripting context is inside another browser extension. I am out of time to hack on this, but I think a good next step is to see whether you can [assert that an injected iframe has the expected content](http://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Matchers) since it will tell you whether Selenium is getting blocked by everything in the extension context. The steps are:

1. Create a web page that has 1 iframe and 1 link. The iframe should have a source document that has the text "it works" and a URL served from http. The link should have a URL that will be injected by the safari extension.
1. Write a test that [expects](http://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Matchers) the iframes to have the "it works" text. You will need to change capybara to each iframe's window in order to complete this test.

You need to have the http iframe so you know your test should work on the safari iframe. If this test case doesn't work then we will want to show it to the Selenium developers and open a bug report. Getting good selenium tests is extremely important for maintaining the project beyond the end of the summer.

sammyshj:
@smcgregor Hello,
Yes, I have tried out commenting some tests in `ts_all_specs.rb` file so as to run specific tests. It looks like the tests which do not involve login run successfully. The tests that involve logging in fails.
The `page.driver.browser.get` works for other URLs but hangs when it is directed to any URL of the extension.
As you pointed out, Safari is rejecting the Selenium WebDriver browser extension when the scripting context is inside another browser extension.
Also, as you mentioned in your comment, I will see whether I can assert an injected frame has the expected content. This will follow the 2 points that you have mentioned.
Thanks for helping me out here

sammyshj:
@smcgregor As you mentioned that a new web page needs to be added. The `page.driver.browser.get` does not navigate successfully to the privly safari extension web pages. However, the page navigates to any URL successfully other than the extension web pages itself.
I will update this PR with a new commit shortly that shows the same.

sammyshj:
@smcgregor I have added a new commit https://github.com/sammyshj/privly-applications/commit/ec5894da3ef9bd24df62a81dda048c10e2c960bc
In this, the test first acquires the safari extension URL and then visits the external link http://web.iiit.ac.in/~sambuddha.basu/privly_safari_test.html to inject the privly application.

smcgregor:
More details please. Does this pass?

I don't think this answers whether WebDriver is able to query the contents of an iframe whose src is the extension.

sammyshj:
@smcgregor Yes, the current commit test passes successfully.
However, if the URL of the test page is changed to that of a local extension page in https://github.com/sammyshj/privly-applications/blob/ec5894da3ef9bd24df62a81dda048c10e2c960bc/test/selenium/specs/tc_safari.rb#L16, the test does not pass then and hangs at that exact line.

smcgregor:
We need to know whether Selenium has the ability to send commands to an iframe if the iframe has a src in the extension context. Your test currently checks whether you can examine the contents of the http src'd iframe, but I don't think it is necessarily trying to get into the extension's iframe. Can Selenium see the contents of the extension's iframe?

If it can see the contents of the extension iframe then it is possible to run many different tests for the extension context, otherwise we need to send a bug report to the Selenium people.

sammyshj:
@smcgregor Can you explain what exactly is meant by the extension iframe in the above comment? Also, I did not understand completely what is meant by the iframe having a src in the extension context.

smcgregor:
If the iframe is served by the extension, it has a src pointing to safari-extension://

sammyshj:
@smcgregor That is already being tested in the previous commit. It can be seen at https://github.com/sammyshj/privly-applications/blob/ec5894da3ef9bd24df62a81dda048c10e2c960bc/test/selenium/specs/tc_safari.rb#L18
The iframe src is tested to be pointing to `Capybara.app_host`. In the case of the Safari extension, the `Capybara.app_host` stores the base URL of the Safari extension.

smcgregor:
I think I see the problem now. The src attribute is part of the http page, not the safari-extension page. You need to attempt to query the actual contents of the iframe.

sammyshj:
@smcgregor Do you mean that the current webpage http://web.iiit.ac.in/~sambuddha.basu/privly_safari_test.html should be included as a local webpage inside the extension and then navigated to from the test?

smcgregor:
No, I mean that the iframe that the extension injects has its own HTML document, and you need to see if selenium can access it.

sammyshj:
@smcgregor In commit https://github.com/sammyshj/privly-applications/commit/4809d2b195cae03d84760b806e3033c819188b68, the iframe contents are being tested. However the iframe tested is a static iframe.
The iframe that is injected by the safari extension is done after some delay(this includes loading the application). I am currently working on adding a delay to the tests to wait for the extension injected iframe to be tested.

sammyshj:
I think the test cannot switch to the injected iframe as the iframe is not present right when the webpage is loaded and is injected later. What do you think about this?

sammyshj:
I think the test cannot switch to the injected iframe as the iframe is not present right when the webpage is loaded and is injected later. What do you think about this?
However, the iframes that are already present in the webpage during load, can be accessed and their contents can be tested on Safari.

smcgregor:
The purpose of Capybara is to wait for elements to load into the page before proceeding with tests. This should include waiting for things like iframe to load.

sammyshj:
I think I need to discuss the same issue with the Capybara and Selenium developers too. I will check if I find any responses on their IRC channels.
This is because, in Safari, whenever I try to navigate to the application pages, the pages do not load completely and hang.

sammyshj:
@smcgregor I have tried adding other implicit and/or explicit waits but they don't seem to help too. Hence, I have written a post on Capybara's mailing list, which can be seen at, https://groups.google.com/forum/#!topic/ruby-capybara/hVgTd-dDi6c
Do let me know, what should be done about the integration tests for the Safari extension.
At this point of time, I think the best way out would be to wait to get a reply from the Capybara community and go ahead with the implementation of the other things left and the unit tests. In the meantime, I will see if any other tests work.
The main problem seems to be that when navigated to a Safari extension page, Capybara does not seem to identify that the webpage has been loaded completely.
Let me know your thoughts on this matter.
Thanks

smcgregor:
I think the problem here is with Selenium's Safari driver, not with Capybara. When you attempt to control the page directly with the Selenium driver, the problem is still present.

This is turning into a black hole, let's keep working on implementation but contact the Selenium people?

sammyshj:
@smcgregor Yes, I will write a mail about the same on the Selenium mailing list

smcgregor:
@sammyshj Should I attempt to merge this with the gsoc-develop branch?

sammyshj:
@smcgregor That is a good idea. Im not sure whether the tests will pass but, its worth having a try. Thanks!

smcgregor:
@sammyshj I think some of your current work is related to fixing issues with this pull request. Do you want me to merge this into gsoc-develop and we'll figure out what is happening with SauceLabs?

sammyshj:
I think I will work on this a bit more and then it can be merged with gsoc-develop.

smcgregor:
Sounds good, ping me when ready.

sammyshj:
@smcgregor This PR was initially regarding running the integration tests for the Safari extension on SauceLabs. But, it seemed like it is not possible to run the extension tests as the pages would not load completely using Selenium.
One more thing that happened was that the Selenium tests without the extension had started to fail after upgrading the version of Safari on SauceLabs. So I think the PR title should be changed to the same and changes should be made on getting the tests to run properly without the extension?

smcgregor:
There is some value in SauceLabs testing for the extension since we can still detect when an iframe injects, but I don't think this single test case is worth the development effort. Yes, let's focus on completing this pull request for running tests on SauceLabs without the extension.

sammyshj:
This PR is for merging the code into the 'develop' branch. So, I think that this PR can be closed and then, I will open a new PR to merge the code to the 'gsoc-develop' branch.

PR #262, opened 2015-06-22
sauce.json added in gitignore

sammyshj:
Added sauce.json in .gitignore.
The sauce.json file contains information about username and accessKey for connecting to SauceLabs. This is different for every user and hence would be appropriate to be present in gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 *.swp
 .idea
 *~
+test/sauce.json
 test/node_modules/*
 test/coverage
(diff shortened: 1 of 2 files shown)


smcgregor:
It is already in the git history, won't this still try to commit changes?

sammyshj:
@smcgregor I have removed the sauce.json file in commit, https://github.com/sammyshj/privly-applications/commit/4309c77094477aa300d1e102e9076eef55401e70
The sauce.json file gets created when npm install is run in the test directory. This is because, the node install.js command is executed which then creates the sauce.json file, https://github.com/privly/privly-applications/blob/master/test/install.js

sammyshj:
Should I squash both the commits now?

PR #263, opened 2015-06-23
Coverage files are test specific now

sammyshj:
If the karma.conf-ci.js file is not run from the same directory, this means that the file has been run with the Safari run_each.sh script. In this case, we want to load coverage files under Safari extension's scripts directory only, https://github.com/privly/privly-safari/tree/master/privly.safariextension/scripts

### DIFF
--- a/test/karma.conf-ci.js
+++ b/test/karma.conf-ci.js
@@ -3,8 +3,22 @@ var uuid = require('uuid');
 
 module.exports = function(config) {
 
+  // These are the default coverageFiles. They should be updated when run from
+  // the privly-safari repo, https://github.com/privly/privly-safari
+  var coverageFiles = {
+    '*/j*/!(raw*).js': 'coverage'
+  };
+
   // If the script is not being executed from the testing directory
   if(process.cwd() !== __dirname) {
+
+    if(process.cwd().indexOf("privly-safari") != -1) {
+      // When the script is run from the privly-safari repo, update the coverageFiles
+      coverageFiles = {
+        '../scripts/*.js': 'coverage'
+      };
+    }
+
     console.warn(
       "\n!!!\nYou are running this script from outside the test directory. " +
       "If you do not have the required node modules on your NODE_PATH, " +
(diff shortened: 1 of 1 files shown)


test/karma.conf-ci.js
```diff
@@ -3,8 +3,15 @@ var uuid = require('uuid');
...
 module.exports = function(config) {
 
+  coverageFiles = {
+    '*/j*/!(raw*).js': 'coverage'
+  };
+
   // If the script is not being executed from the testing directory
   if(process.cwd() !== __dirname) {
+    coverageFiles = {
+      '../scripts/*.js': 'coverage'
```
smcgregor: I like this approach more than my other suggestion, but you could improve it by getting the path to the directory and checking for privly-safari. You should also leave a comment above this block pointing to the Safari repository.

sammyshj:
@smcgregor Made the necessary changes and squashed the commits into one.

test/karma.conf-ci.js
```diff
@@ -3,8 +3,21 @@ var uuid = require('uuid');
 
 module.exports = function(config) {
 
+  coverageFiles = {
```
smcgregor: The comment explaining this should be up here now.

sammyshj:
@smcgregor Fixed the comments

test/karma.conf-ci.js
```diff
@@ -3,8 +3,22 @@ var uuid = require('uuid');
 
 module.exports = function(config) {
 
+  // These are the default coverageFiles. They should be updated when run from
+  // the privly-safari repo, https://github.com/privly/privly-safari
+  coverageFiles = {
```
smcgregor: Whoops, missed this before, but I believe this should be `var coverageFiles = {`

sammyshj: Yes, thanks for pointing that out!

sammyshj:
@smcgregor Made the small change. The PR can be merged now.

PR #266, opened 2015-07-10
Added necessary javascript include files

sammyshj:
**The changes done in the PR:**
1. Included the javascript files `context_messenger.js`, `storage.js`, `glyph.js`, `options.js` to the privly-application pages
2. Removed the previously included javascript file `extension_integration.js` from the privly-application pages
3. The above mentioned privly-application pages include:
* Help/new.html
* History/new.html
* Login/new.html
* Message/new.html
* Message/show.html
* Pages/ChromeOptions.html
* PlainPost/new.html
* PlainPost/show.html

In `network_service.js`, the `platformName` function now returns `"SAFARI"` when run from the Safari extension.
@smcgregor The changes have been verified with the Safari extension and they work good now.

### DIFF
--- a/templates/base.html.template
+++ b/templates/base.html.template
@@ -14,6 +14,10 @@
 
     <script type="text/javascript" src="../shared/javascripts/meta_loader.js"></script>
     <script type="text/javascript" src="../shared/javascripts/parameters.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/context_messenger.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/glyph.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/storage.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/options.js"></script>
     <script type="text/javascript" src="../shared/javascripts/network_service.js"></script>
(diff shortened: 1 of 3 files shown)


smcgregor:
You should change the .template and .subtemplate files and leave the .html files alone. I believe all the HTML files are templated.

sammyshj:
I have now changed only the .template file now. This file is included in all the privly-application pages that has been mentioned in my previous comment.
Presently, as only the .template file is changed, I think we will need to run the build.py script to make the necessary changes in the .html file.

templates/base.html.template
```diff
@@ -14,6 +14,10 @@
 
     <script type="text/javascript" src="../shared/javascripts/meta_loader.js"></script>
     <script type="text/javascript" src="../shared/javascripts/parameters.js"></script>
+    <script type="text/javascript" src="../shared/javascripts/context_messenger.js"></script>
```
smcgregor: I'm surprised none of these were referenced in the template previously.

sammyshj:
I think the .html files need to be updated from the .subtemplate by running the build.py script.

smcgregor:
I generally only update the rendered html files when updating an extension/web server's git submodule.

PR #269, opened 2015-07-13
Allow reuse of LocalStorage shim

hitesh96db:


smcgregor:
The commenting needs some work. The "case 1," "case 2" labels are not helpful. You should also write bit more about what simple storage is (for people not developing the Jetpack extension), the differences between the module version and the preferences version, and indicate that the preferences version is deprecated.

hitesh96db:
Oops, my bad, the comments are misleading, `simple-storage` isn't used anymore.

hitesh96db:
Changed the comments.

shared/javascripts/local_storage.js
```diff
@@ -87,14 +87,22 @@ var ls = {
   }
 };
 
-// Determine whether localstorage can be used directly
```
smcgregor: The try block should have a comment indicating that it is being used to test which platform it is running on and selecting the appropriate storage driver.

hitesh96db: Done.

shared/javascripts/local_storage.js
```diff
@@ -87,14 +87,24 @@ var ls = {
...
-                         .getService(Components.interfaces.nsIPrefService)
-                         .getBranch("extensions.privly.");
+  // Firefox
+  // Use Preferences service instead of localStorage
+  if (typeof module !== 'undefined' && module.exports) {
+    // Loaded as a CommonJS module, reused as a Local Storage shim 
+    // in lib/local_storage.js.
+    module.exports.ls = ls;   
+  } else {
+    // Privileged JS code run in a chrome page. for eg:- ChromeFirstRun page.
```
smcgregor: Unfortunately since Google decided to name their browser "chrome," you should be more specific when you say "chrome page".

smcgregor: Otherwise, this is ready to merge.

hitesh96db: Done.

PR #272, opened 2015-07-17
Fix SafariAdapter message format

sammyshj:
Changes made in the PR:
* In each of the `dispatchMessage()` functions, we need to use `name` and a `message` keys.
Example: If the function is called as `dispatchMessage(a, b)`, then, `a` is `name` and, `b` is `message`
* While we receive the messages, we need to check if the message was sent using `dispatchMessage()` function. If done so, then, the information is contained in the `payload.message` part and thus we do a `callback(payload.message)`. If the message has been sent by `postMessage()` then, the information is the whole `payload` and then, in this case, we need to do, `callback(payload)`.

@smcgregor @SummerWish In the Safari extension, when the `PRIVLY_APPLICATION` pages are opened, there seems to be a different `localStorage` from the one of the background page. Hence, whenever an option is changed, and the `messageExtension()` and `messageContentScript()` is called, the localStorage values need to be updated in the background page localStorage too.
I think Chrome shares the same localStorage for both the extension pages and the background page. Let me know if I am thinking right on this topic.

### DIFF
--- a/shared/javascripts/context_messenger.js
+++ b/shared/javascripts/context_messenger.js
@@ -439,7 +439,7 @@ if (Privly === undefined) {
         safari.self.contentWindow.postMessage(payload, '*');
       }
       else {
-        safari.self.tab.dispatchMessage(payload);
+        safari.self.tab.dispatchMessage("privlyMessage", payload);
       }
       return;
     }
@@ -452,7 +452,7 @@ if (Privly === undefined) {
         safari.application.activeBrowserWindow.tabs.forEach(function (tab) {
           // Don't message Privly Applications
           if (tab.url.indexOf('safari-extension') !== 0) {
-            tab.page.dispatchMessage(payload);
+            tab.page.dispatchMessage("privlyMessage", payload);
           }
         });
       }
@@ -461,7 +461,7 @@ if (Privly === undefined) {
     if (to === 'PRIVLY_APPLICATION') {
       // Send message to all content scripts
       safari.application.activeBrowserWindow.tabs.forEach(function (tab) {
-        tab.page.dispatchMessage(payload);
+        tab.page.dispatchMessage("privlyMessage", payload);
       });
       return;
     }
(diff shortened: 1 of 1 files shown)


SummerWish:
Yup. Chrome shares localStorage

sammyshj:
@smcgregor Added a new commit, https://github.com/sammyshj/privly-applications/commit/256c209bc9bec015f2717d7922f24ebe8ee98d58
**Changes made in the new commit:**
In the Safari extension, the `localStorage` is different for the `PRIVLY_APPLICATION` pages and the `BACKGROUND_SCRIPT`. Hence, when an `options/changed` message is received by the `BACKGROUND_SCRIPT`, the `localStorage` values are now updated in the background page also.
Option values that are updated are:
* options/privlyButton
* options/injection
* options/whitelist/domains
* options/whitelist/regexp
* options/contentServer/url
* options/glyph

shared/javascripts/context_messenger.js
```diff
@@ -471,12 +471,39 @@ if (Privly === undefined) {
...
       safari.application.addEventListener("message", function(payload) {
-        callback(payload);
+        if (payload.name !== "undefined" && payload.name === "privlyMessage") {
+          callback(payload.message);
+          payload = payload.message;
+        } else {
+          callback(payload);
+        }
+        // Update the options/changed values in the background page
+        if (payload.body !== "undefined" && payload.body.action === "options/changed") {
```
smcgregor: I believe these should be set as listeners in the background page, not in the messenger script.

sammyshj: @smcgregor Does that mean adding these lines to the privly-safari repo?

smcgregor: So all these lines are to ensure the background script has the same local storage options as the privly-app? It would be better to change the local storage script to send/receive a storage message for the safari platform that will ensure the two scripting contexts store all the same values. This will be less brittle.

sammyshj: I have updated the PR to have only the initial commit now. If you think the code looks good, then the PR can be merged.

sammyshj:
@smcgregor Can this PR be merged, or does it require any other changes?

shared/javascripts/context_messenger.js
```diff
@@ -471,12 +471,20 @@ if (Privly === undefined) {
   SafariAdapter.prototype.setListener = function (callback) {
     if (this.getContextName() === 'BACKGROUND_SCRIPT') {
       safari.application.addEventListener("message", function(payload) {
-        callback(payload);
+        if (payload.name !== "undefined" && payload.name === "privlyMessage") {
```
smcgregor: Does this need to be `typeof payload.name !== "undefined"`?

sammyshj: Yes, updated the PR.

shared/javascripts/context_messenger.js
```diff
@@ -471,12 +471,20 @@ if (Privly === undefined) {
...
+        }
       }, true);
     }
     if (this.getContextName() === 'CONTENT_SCRIPT') {
       safari.self.addEventListener("message", function(payload) {
-        callback(payload);
+        if (typeof payload.name !== "undefined" && payload.name === "privlyMessage") {
+          callback(payload.message);
+        } else {
+          callback(payload);
```
smcgregor: What cases require this else statement?

sammyshj: The else statement is called when the message is sent from a `postMessage()` function.

smcgregor: Ok, it needs comments indicating the scripting environments that will receive messages at that branch of the `if` statement, and which scripting environments will send messages to that branch of the `if` statement.

sammyshj: Added the comments now at https://github.com/privly/privly-applications/commit/f651eb59418ae8c2298db249296b9da804d7b286#diff-e3c7938b324cb640951cd58dd4d34be2R475

PR #270, opened 2015-07-14
Seamless posting

SummerWish:

PR #274, opened 2015-07-19
Seamless-posting

SummerWish:
All of the commits related to seamless-posting are now merged into a few, so you should be very comfortable to review changes.
This PR is different to #270 :
- `embeded` is renamed to `seamless-posting` to avoid confusion, no more `embeded` now
- refactored the `new` action and adapted the message App and plainpost App to it. Minimal changes were made to history/login/help Apps: I think they need rewrite.
- generated HTML are not included, in order to keep the diff clean and easy to review
- tests are not included currently

Merging commits is HORRIBLE:
- If I forget to pick up something in a commit, then all works after that commit will be reverted, in order to insert the change into that commit.
- I cannot compare automatically to see if I missed picking up something: file names are changed, variable names are changed.. a lot of things are changed. branch diff isn't smart enough to detect those changes. The diff result is junk.
- SURPRISE! This branch seems to work!


Message/js/seamless.js
```diff
@@ -0,0 +1,13 @@
+/*global Privly */
```
smcgregor: Needs file overview.

Message/js/seamless_ttlselect.js
```diff
@@ -0,0 +1,6 @@
+/*global Privly */
```
smcgregor: Needs @fileoverview

smcgregor:
Thanks for breaking things up. Do you generate test coverage with coveralls.io? First step is to make sure we aren't decreasing unit test coverage.

Pages/SeamlessPostingLoginCallback.html.subtemplate
```diff
@@ -0,0 +1,7 @@
+{% extends "templates/base.html.template" %}
+
+{% block head %}
+  <script>
+    window.close();
```
smcgregor: You can't put inline scripts in the chrome extension context.

When logging in from the seamless posting button, the login window redirects here and the window doesn't close.

smcgregor:
* When I click outside the seamless posting form, the posting icon in the chrome context does not change color from green to red in all cases. This means the host page will be able to spoof the posting form. Further, a large proportion of the population is red/green colorblind, which makes people unable to differentiate the red icon from the green icon. I recommend changing the icon to:

green = the user's input will be captured by a chrome context page
gray = the user's content will not be captured by the chrome context page

* When the seamless iframe is small (see [case 3](http://test.privly.org/test_pages/embedposting_iframes.html)), the ability to set the expiration time gets truncated. I don't think we need so many options, how about only the three options 1, 7, and infinite.

* You removed the context menu?

* When posting a status update on FB, everything works as well as one could expect, but immediately after posting another seamless posting form gets injected into the status update box. I would guess you needed to do this to solve the strange behavior of FB's forms, but do you have an easy way to prevent this from happening?

smcgregor:
[There are numerous failing unit tests](https://travis-ci.org/smcgregor/privly-applications/builds/72519184).

SummerWish:
I just added documentations as comments.

I want to point out that, I refactored (part of) the architecture of a Privly App in this PR, which you may need to take a look especially.

For convenience, I just paste some documentations in the commit here, thus you can learn about the new architecture introduced in this PR:

```js
/**
 * About the MVC architecture of the Privly application:
 * 
 *   An interactive Privly application can be splitted into
 *   three parts: Model, View, Controller.
 *
 *   First of all, let's determine different use cases of a
 *   Privly app. Generally, a Privly app has three use cases:
 *   
 *   1) show
 *      When the link is embedded into a host page, this
 *      application will be embedded to the host page to present
 *      the content of the link. This is the `show` case.
 *      
 *   2) new
 *      When user clicks `new` button in a browser action button,
 *      or in a context menu, or in the Privly navigation bar, this
 *      application will be shown to user, allowing user to create
 *      such kind of content. In this case, it is `new`.
 *      About editing an existing content, it should be categorized
 *      to `new` case. However due to historical issue (?), it is
 *      now charged by the `show` case.
 *
 *   3) seamless
 *      We introduced a new method of creating content: seamless-
 *      posting. For `new` case, the users are creating content in
 *      a saparated, standalone window, and the content is created
 *      when users click 'Save' button. However in `seamless` case,
 *      the users are creating content seamlessly in the host page.
 *      Besides, the content is created once users enter `seamless`
 *      mode, and will be frequently updated or finally deleted.
 *      In short, in `seamless` case, the user is going to create
 *      the content using the Privly App, just inside the host page.
 *
 *   Now let's see what are the MVC of a Privly application.
 *
 *   VIEW LAYER
 *   ==========
 * 
 *   The view layer is the user interface layer. Currently we have
 *   three view prototypes: show, new and seamless. Yes as you can
 *   see, these three view prototypes provide templates for the 
 *   three use cases. We offer templates for App developers, to
 *   allow faster App development. However those templates are
 *   free to be inherited, extended or overridden so it is still
 *   flexable. For example, our `show` view prototype provides a
 *   textarea for users to input contents. However you can override
 *   that part of the template to allow users dragging image into
 *   it.
 *
 *   Notice that, if your App supports `seamless` view, it should
 *   also provides a `seamless_ttlselect` view, used to display
 *   seconds_until_burn menu for user. If you don't want to change
 *   menu styles, or change menu orders, using our default
 *   prototype is enough for your App.
 *
 * 
 *   MODEL LAYER
 *   ===========
 *   
 *   The model layer handles such basic tasks below:
 *   
 *   - How should the user inputted content in the view layer be
 *     transformed into a Privly structured_content? This task is
 *     used in publishing or editing the content.
 *
 *     Example:
 *     For Plainpost App, no transformations.
 *     For Message App, the transformation is encryption.
 *     For Image App (not existed yet), the transformation is
 *     binary serialization and encryption.
 *     
 *   - How should we transform a Privly structured_content to the
 *     content that the view layer can present? This task is just
 *     a reserve situation of the task above. It is used when user
 *     is going to see the content (`show` case).
 *
 *     For example, for Message App, the transformation is decryption
 *     according to the encryption key attached in the link.
 *
 *   - What are the `seconds_until_burn` options for this App? An
 *     app can assign its own seconds_until_burn (aka. TTL, Time
 *     To Live) option items.
 *
 *   Notice that the model part can provide other interfaces as well.
 *   Generally those interfaces are useful in the controller layer.
 *
 *   Also please notice that, generally all interfaces of the model
 *   layer should be async, by returning a Promise which will later
 *   be resolved. The async feature of a model doesn't have much
 *   effect in current Privly Application like Message and Plainpost,
 *   however it does offer developers more flexibility.
 *   
 * 
 *   CONTROLLER LAYER
 *   ================
 *
 *   Controller layer connects model layer and view layer. It can
 *   offer extra actions for every view layer events, for example,
 *   for a Message App, when a new link is created, we have to concat
 *   the encryption key after the link. Another example is, when
 *   we are presenting the content of a Message App, we should do a
 *   Markdown rendering. This task should also be completed by the
 *   controller layer.
 *
 *   In the Privly application, you needn't explicitly connects
 *   all interface between view and model. You can just pass the
 *   model to the view! The view will call related interface
 *   itself. For a controller, you only need to write initializing
 *   code and write event handlers if you want to change the
 *   default process of a prototype view.
 *   
 * For Message App, the controller layer for each view is:
 *   new:       Message/js/new.js
 *   show:      Message/js/show.js
 *   seamless:  Markdown/js/seamless.js
 *   seamless_ttl: Markdown/js/seamless_ttl.js
 * 
 */
```

- How do you think of this new architecture?

- As pointed out in the documentation, I think "editing a content" should be adapted to the `new` use-case (currently it is `show`), what's your idea?

- Under the new architecture, I think we can introduce one more view prototype: `standalone`. This view is especially designed for login, options and history applications. They do not have a `new` process however they do need to present contents in a saparated and standalone window (tab) like a `new` process. Thus providing a more generic view prototype for those applications can make the structure better.

  standalone: a kind of Application view that is rendered in a standalone tab

  new: based on `standalone`, provides a basic textbox and publish button, for the `New xxx` action.

  show: a kind of Application view that is presented when it is injected into the host page, or when user is directly accessing the link.

  seamless: view of the seamless-posting environment.

smcgregor:
I think the general outline of your changes is good, and it would help organize things that have been developing for a while. However, I am concerned about refactoring so much as we are coming up on the last phase of the summer timeline. Can you get (at least) the 70 percent code coverage we had previously on unit tests? Otherwise, I can't merge it. We can't backslide on test coverage.

If it is realistic to get to 90 percent coverage (with 70 percent as the minimum), then I will dive into your proposal and make suggestions as necessary.

SummerWish:
> When the seamless iframe is small (see case 3), the ability to set the expiration time gets truncated. I don't think we need so many options, how about only the three options 1, 7, and infinite.

I think a better solution is to move the menu into the top most iframe, just like embedded-posting does. However I need to find solutions to contact the top most iframe first. currently the message mechanism doesn't allow developer doing this.

SummerWish:
> You removed the context menu?

Yes, temporarily. This might require some architecture change so I postponed it. Before the end of summer it will be fixed

smcgregor:
[I ran this on TravisCI](https://travis-ci.org/smcgregor/privly-applications/builds/73084113) and there are a good number of failures, some of them are on Chrome. [Here is your coverage data](https://coveralls.io/builds/3163253).

SummerWish:
Sorry it's my fault. I changed `run_each.sh` to add essential dependency for testing `seamlessPosting` but forgot to commit it. I have pushed it now and there should be no errors when running unit test on Chrome.

smcgregor:
[Running again](https://travis-ci.org/smcgregor/privly-applications/builds/73257995).

smcgregor:
Coverage is currently [67.8](https://coveralls.io/builds/3172795), so coverage is still lower.

What is happening with these requests in the build:

    WARN [web-server]: 404: /top-expect-404.css
    WARN [web-server]: 404: /top-expect-404.js
    WARN [web-server]: 404: /injected-expect-404.css
    WARN [web-server]: 404: /injected-expect-404.js

SummerWish:
They are here:
https://github.com/SummerWish/privly-applications/blob/seamless-posting/shared/test/meta_loader.js

SummerWish:
Some issues with this coverage report..
- Coverage of view adapters are not included, which will definitely make the coverage higher if it was included.
- It seems that, the lower average coverage is caused by my simulation to the `shared/javascripts/network_service.js` interface in new test cases. Essentially speaking, we got a lower coverage because those part of the code are just not tested formerly. I removed dependencies of calling such kind of code in new test cases, which makes the coverage lower. I think this is not an testing issue introduced in this PR.

irdan:
Regarding the editing of existing content, I think it would be more intuitive to have the `show` use case be distinct from the `new` use case. Instead of lumping the edit functionality into `show`, I suggest creating an extension of `new` called `edit`.  This way `show`, `new`, and `edit` each have a single responsibility.  

It's a little more work right now, but it is a lot saner from a design standpoint and wouldn't require every future developer to learn or remember that editing is performed in show.

Message/js/messageApp.js
```diff
@@ -0,0 +1,312 @@
+/**
+ * @fileOverview This file provides the model layer of Message
+ * App. Message App is a simple Privly application that supports
+ * text encryption.
+ *
+ * Fore more details about the Message App, please visit Privly
```
irdan: nit: s/Fore/For/

smcgregor:
I don't think we should necessarily add additional view types immediately since the injected "show" endpoint is also the "edit" endpoint. There is going to be confusion regardless.

irdan:
Could the edit/show issue be addressed by having multiple controllers for the single `show` view? Admittedly I don't have experience with MVC in the context of a web app, so I don't know if this is against best practices.

smcgregor:
I think we can compare privly apps to the MVC design pattern, but we can't directly invoke it. Now that we know how everything can fit together among all the platforms, we could use a "design pattern" doc, but I think that is better to land at v2 and port the existing style first.

SummerWish:
The MVC introduced in this PR is in fact a saparation of the model and the view. I was not intended to introduce MVC at the beginning, only wanted to extract a layer which does not involve with any DOM operations and any HTML status, only cares the foundation operations, and could be reuseable across different templates. After I completed extracting, I found that the new architecture is very similiar to the MVC pattern so I use MVC as a name here.

smcgregor:
With Rails experience, calling it MVC is confusing, but [naming things is hard](http://martinfowler.com/bliki/TwoHardThings.html). I'll see if I can come up with a better name. In the meantime, this is no reason to slow development. I will inwardly think of this as the ["new hotness"](https://youtu.be/ha-uagjJQ9k?t=17s)

SummerWish:
I think the core idea of MVC is the separation of operations(model) and user interfaces(view) and using controllers to connect them. After saparation, the presentation and data manipulating are no longer mixed together. What the controller is doesn't matter since as long as we defined model and view, we will know what's the controller. controller just connects them. There always should be something connects the two layers, that's the controller. In this way, common operations are reuseable since they are encapsuled.

This is a kind of design pattern, and as far as I know MVC design pattern matches it best because the role of each part just naturally matches it.. No intention, it just matches.

SummerWish:
@irdan I think the story of abstract more is complex..
Currently the model is shared, but templates are not shared. We only need to write once to transform a user inputted content into a Privly json, that's good. But considering a new kind of application that takes images as contents. In the view layer we still need to handle image dragging upload in "edit" view and "seamless" view. Thus we still need to write twice, in order to handle "dragging".

The most ideal situation is, we not only share models, but also share templates. For a template, we only have three different kinds: edit, show and normal (which does not involves with data models, like login).

```
+ standalone (directly access)
  - view
  - edit
+ embedded (embedded access)
  - view (seamless)
  - edit (seamless)
```

I do not have a clear idea to achieve "template code" sharing currently, thus I only completed "model" sharing. Maybe when I start to implement the mixed-content Privly application I can figure out a way to share template.

smcgregor:
In looking over the files changed, I think I need to see more organization enforcing the separation between model, view, and controller. In Rails these components are separated into different directories. Since we want the views to be at the root of the application's directory, the controllers and models will likely need to be in their own sub directories.

In the old version there was little inheritance, which made it easy to figure out how everything plugged together by viewing the files. I think with the changes you are introducing we need to add more documentation to [priv.ly/develop](https://priv.ly/pages/develop) that explains how each of these elements fit together. 

One of the reasons I am resisting "MVC" as a label is I am am not clear on the separation between the model/view/controller. Much of the confusion is coming from my Rails programming experience, so many other devs will likely be similarly confused. Are your Model/View/Controllers all Javascript files? In Rails the view is in a templating language. However, since your controllers are not aware of the DOM they apply operations to, I don't see how the view can be in a templating language or in the HTML.

Maybe a stupid question, what does ttl stand for?

smcgregor:
todo: work way through each of the shared scripts and review them for style, coverage, and functionality

## Shared Javascripts

event_emitter.js

* Style and test coverage look good.
* I don't know why you created the shared javascript. Can you add details to the file interview that indicate what you use the file for? It looks like some boilerplate code you use, but I don't know its purpose.

context_mesenger.js

* Change [line 245](https://coveralls.io/builds/3199504/source?filename=shared%2Fjavascripts%2Fcontext_messenger.js#L245) to `window.location.href.indexOf("chrome://" + window.location.host + "/privly-applications/") === 0`
* Does the content script forward PRIVLY_APPLICATION messages to the injected privly apps? If, I don't see where this happens. I don't believe anything uses this at the moment, but it would be good to know.
* Coverage needs to improve. You are responsible for the Chrome coverage, while others are responsible for the Safari/Firefox/etc coverage. You will need to mock a few Chrome APIs for this to work.
* Implementation looks solid.

glyph.js

* [This function needs coverage](https://coveralls.io/builds/3199504/source?filename=shared%2Fjavascripts%2Fglyph.js#L182).
* Is there a reason you implemented the canvas version of the glyph, aside from it being a more consistent display? It seems like we should deprecate the table version.

host_page_integration.js

* Good.

local_storage.js

* Good.

meta_loader.js

* Good.

network_service.js

* Good.
* I think most of this code will be replaced by by "models" under your proposed changes? No need to change it for now.

options.js

* Good/reviewed previously.

parameters.js

* [This needs coverage](https://coveralls.io/builds/3199504/source?filename=shared%2Fjavascripts%2Fparameters.js#L90), but it is my responsibility.

storage.js

* [this](https://coveralls.io/builds/3199504/source?filename=shared%2Fjavascripts%2Fstorage.js#L65) should always return `undefined` for this to mirror the local storage API. The synchronous storage should mirror local storage as closely as possible.

tooltip.js

* Coverage is bad, but I think this is my problem.

### tests

* [update this](https://github.com/SummerWish/privly-applications/blob/c27526dceefa644512f0367b32299d46ea3b3f56/shared/test/context_messenger.js#L42) and [this](https://github.com/SummerWish/privly-applications/blob/c27526dceefa644512f0367b32299d46ea3b3f56/shared/test/context_messenger.js#L65). It is critical that we have strong tests for establishing trust in origins.
* The [tooltip.js suite](https://github.com/SummerWish/privly-applications/blob/c27526dceefa644512f0367b32299d46ea3b3f56/shared/test/tooltip.js) needs updating, but this is my problem.
* The adapters in shared/privly-web are currently not generating coverage information. The regular expression [here](https://github.com/SummerWish/privly-applications/blob/c27526dceefa644512f0367b32299d46ea3b3f56/test/karma.conf-ci.js#L9) needs updating. I can do this if you want.

## Pages

* What is [this file](https://github.com/SummerWish/privly-applications/blob/c27526dceefa644512f0367b32299d46ea3b3f56/Pages/js/message_test.js)? It needs a file overview and it should probably be in the test subdirectory of its current directory.
* [This needs a @fileOverview](https://github.com/SummerWish/privly-applications/blob/c27526dceefa644512f0367b32299d46ea3b3f56/Pages/js/seamless_login_callback.js) saying why it exists and what uses it. No exceptions.
* Since I am not getting coverage information for options.js and it is found in [runEach.sh](https://github.com/SummerWish/privly-applications/blob/c27526dceefa644512f0367b32299d46ea3b3f56/test/run_each.sh), I think there is something wrong with the coverage data on non-shared JS.
* Move [this file](https://github.com/SummerWish/privly-applications/blob/c27526dceefa644512f0367b32299d46ea3b3f56/Pages/MessageTest.html.subtemplate) into a testing directory and update its manifest for the build.

## General Questions

* What is the meaning of ttl? We might need to expand this since it is not immediately obvious.
* Which files represent controllers, views, and models are not obvious. They need to be better separated, probably in their own directories.

## General Comments

Things are looking _very_ good. You refactored a ton of code and I don't have any huge issues, which is impressive. We just need to get it merged ASAP because the gsoc-develop branch is going to be changing underneath this. I'll review the Message and PlainPost app changes after it is clear where your model/view/controllers are.

SummerWish:
> event_emitter.js: I don't know why you created the shared javascript. Can you add details to the file interview that indicate what you use the file for? It looks like some boilerplate code you use, but I don't know its purpose.

For emitting events, it is used in Privly App models such as `shared/javascripts/privly-web/new.js` and `shared/javascripts/privly-web/seamless.js`. For samples of listening emitted events, see those Privly App controllers, such as `Message/js/new.js`, `Message/js/seamless.js`, etc.

Overwriting callback methods are now replaced by emitting events and listening events which could provide more flexibility and readability. For example, in `new.js`, developers can listen `beforeXXX` and returns `true` in the listener callback to disable calling the original function, or returns nothing to continue calling the original function.

---

> glyph.js: Is there a reason you implemented the canvas version of the glyph, aside from it being a more consistent display? It seems like we should deprecate the table version.

It is originally used in embedded-posting, to switch the icon of [modal_button](https://developer.chrome.com/extensions/browserAction) (it accepts a data URI to use custom icons). However in seamless-posting, as long as we use a green-lock icon, this function is not used any more.

It's a good idea to replace DOM version glyph with canvas version glyph since we needn't CSS anymore.

---

> context_mesenger.js: Coverage needs to improve.

OK I would improve this soon.

---

> context_messenger.js: Does the content script forward PRIVLY_APPLICATION messages to the injected privly apps? If, I don't see where this happens. I don't believe anything uses this at the moment, but it would be good to know.

Yeah, it forwards messages to both top Privly applications and injected Privly applications. This is heavily used in seamless posting (and embedded-posting), for content scripts to communicate with Privly applications. Background script is a bridge between content script and injected Privly applications.

The reason for using context messenger instead of `postMessage` is that, context messenger can send messages across iframes. For example:
```
- TOP
  - A
    - B
      - C
```
content scripts in C may want to contact Privly applications in an iframe in TOP. In such situation, `postMessage` is complex.

---

> What is the meaning of ttl? We might need to expand this since it is not immediately obvious.

It is Time-To-Live. TTL are widely used to represent `time/seconds until deletion/removal/invalidate`. I think it is an equivalent and more familiar name to `seconds until burn` (however it is less obvious). The meaning of TTL is explained in `Message/js/messageApp.js`:

> What are the `seconds_until_burn` options for this App? An
>    app can assign its own seconds_until_burn (aka. TTL, Time
>     To Live) option items.

I use TTL as an acronym in codes, for example, `TTLSelect`, `getTTLOptions`, etc.

SummerWish:
Hi, I have rearranged files to match the MVC architecture. Please have a check. You may need to review new changes at a per-commit basis.

smcgregor:
You can ignore the Safari issue since that is an infrastructure problem, but there appears to be two small issues with the build [found here](https://travis-ci.org/smcgregor/privly-applications/builds/74275927).

shared/javascripts/privly-web/new.js
```diff
@@ -1,266 +0,0 @@
...
-   * @param {function} callback A function to call after the login check
-   * was successfull.
-   */
-  pendingPost: function(callback) {
-    
-    privlyNetworkService.showLoggedInNav();
-
-    // Listen for a message containing the initial content for the form
-    // TODO: no receivers?
-    Privly.message.messageExtension({'ask': 'initialContent'}, true)
```
hitesh96db: This breaks the initial content functionality for the older posting process. What would be the right place to add this back ?

smcgregor: I would go with [the view](https://github.com/privly/privly-applications/blob/gsoc-develop/shared/javascripts/viewAdapters/new.js#L281). @SummerWish any thoughts?

SummerWish: Those codes are removed because responses are not sent anywhere after refaction in privly-chrome.

To add it back, you need to:
1) request for initial content at: https://github.com/privly/privly-applications/pull/274/files#diff-045fb9b2b77dc665412fe50ccef3b6fdR281
2) response for initial content requesting somewhere in the content script.

do you have ideas of point 2?

hitesh96db: I'm not really up-to-date with the refactored privly-applications, so it'll take some time before I can pitch in some ideas.

PR #275, opened 2015-07-19
Load interface scripts as CommonJS modules, refactor Firefox Adapter

hitesh96db:
The changes include --
* Load the interface scripts as CommonJS modules so that they can be used in the jetpack background scripts.
* Added more functionality to the Firefox Adapter.
* Options interface bug fix.
* Changed the check for `Array` data-type in `options.js`.

### DIFF
--- a/shared/javascripts/context_messenger.js
+++ b/shared/javascripts/context_messenger.js
@@ -130,6 +130,11 @@ if (Privly === undefined) {
   Privly.message = {};
   Privly.message.adapter = {};
 
+  // CommonJS Module
+  if (typeof module !== "undefined" && module.exports) {
+    module.exports.message = Privly.message;
+  }
+
   /**
    * A unique id for this context.
    * It will be used to identify the response.
@@ -300,8 +305,8 @@ if (Privly === undefined) {
   /** @inheritdoc */
   FirefoxAdapter.isPlatformMatched = function () {
 
-    validContext = ["BACKGROUND_SCRIPT", "CONTENT_SCRIPT",
-                    "PRIVLY_APPLICATIONS"]
+    var validContext = ["BACKGROUND_SCRIPT", "CONTENT_SCRIPT",
+                        "PRIVLY_APPLICATION"]
     if (validContext.indexOf(FirefoxAdapter.prototype.getContextName()) !== -1) {
       return true;
     }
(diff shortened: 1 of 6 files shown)


shared/javascripts/options.js
```diff
@@ -364,7 +377,8 @@ if (Privly === undefined) {
     if (typeof glyph !== 'object' || glyph === null) {
       throw new Error('invalid argument');
     }
-    if (typeof glyph.color !== 'string' || typeof glyph.cells !== 'object' || glyph.cells === null || glyph.cells.constructor !== Array) {
+    if (typeof glyph.color !== 'string' || typeof glyph.cells !== 'object' || 
+        glyph.cells === null || Object.prototype.toString.call(glyph.cells) !== "[object Array]") {
```
hitesh96db: `glyph.cells.constructor !== Array` will not work for CommonJS modules because the `Array` global object isn't the same one in two different modules/env.

hitesh96db:
The changes in the extension can be found in this [commit](https://github.com/hitesh96db/privly-jetpack/commit/153a98459b7097b39e05d8ae6183300e919c146a) or in the [inbetween](https://github.com/hitesh96db/privly-jetpack/tree/inbetween)(probably not a nice/informative name) branch.

shared/javascripts/options.js
```diff
@@ -54,6 +54,7 @@ var Privly;
 if (Privly === undefined) {
   Privly = {};
 }
+var ls;
```
smcgregor: Why is `ls` here?

shared/javascripts/options.js
smcgregor: Notes like these are usually better left as a comment in the source code.

shared/javascripts/options.js
hitesh96db: right, not needed.

shared/javascripts/options.js
hitesh96db: Done. I've rebased the commits so the changes can't be seen as a separate commit.

hitesh96db:
@smcgregor Hey, I've added a new commit to this PR. Is this Ok to merge ?

smcgregor:
I reviewed the new commits, with one comment. Please respond to the comment and I'll run the unit tests on CI now.

smcgregor:
There are some [syntax errors](https://travis-ci.org/smcgregor/privly-applications/builds/73712805).

hitesh96db:
All the syntax errors are nearly the same thing, but I can't understand what's wrong.

smcgregor:
I am re-running the build just in case it is a non-deterministic error.

smcgregor:
Same failures, I am [running an old build](https://travis-ci.org/smcgregor/privly-applications/builds/67697755) to see if it is still passing.

smcgregor:
The old build works. I think the issue may be the `const` declaration. Specifically, I don't think you can assign into a collection like this across platforms and it is messing with the parser: `const { Cc, Ci }`.

hitesh96db:
Yeah, probably, I can change that..

hitesh96db:
@smcgregor I've added the changes.

smcgregor:
Build is running [here](https://travis-ci.org/smcgregor/privly-applications/builds/73743849). I'll work on merging if it passes.

smcgregor:
It looks like you need to use a different name (see the build).

hitesh96db:
@smcgregor I've pushed the changes.

smcgregor:
[Running the build](https://travis-ci.org/smcgregor/privly-applications/builds/73922600). I had to push a different branch because the history of this PR changed.

shared/javascripts/context_messenger.js
```diff
@@ -316,43 +321,66 @@ if (Privly === undefined) {
...
-    worker.port.on("privly", callback);
+    
+    // Add worker to array of workers
+    pushWorker(worker);
+    // Listen for messages.
+    worker.port.on("PRIVLY_MESSAGE", callback);
+    // the workers array should contain only active workers.
+    // remove the workers, that are inactive or destroyed, from the workers array.
+    worker.on("pageshow", function() { pushWorker(this); });
+    worker.on("pagehide", function() { popWorker(this); });
```
smcgregor: Do you need to remove the workers on hide and detach? This breaks deactivating injection, no?

hitesh96db: Nope, it'll work. The background script doesn't message the content scripts to change the current setting. It instead injects new content scripts in all tabs passing the setting as a content script option, based on which the DOM element is inserted/changed. `privly.js` looks at the DOM element and then decides if it needs to run or not.

smcgregor: I think it would be better to use messaging for this in the future, but that is not a problem for this pull request.

shared/javascripts/options.js
```diff
@@ -63,6 +63,16 @@ if (Privly === undefined) {
   }
   Privly.options = {};
 
+  // CommonJS Module
+  if (typeof module !== "undefined" && module.exports) {
+    module.exports.options = Privly.options;
+    // load dependencies
+    ls = require("./local_storage.js").ls;
+    Privly.storage = require("./storage.js").storage;
```
smcgregor: Won't these already by on the `Privly` object when their scripts are included in the HTML page?

hitesh96db: This is when the script will be loaded as a CommonJS module in the background script. The background script will include a `require("chrome/content/privly-applications/shared/javascripts/options.js")` statement. There isn't any HTML page here.

smcgregor: So the options script coordinates the loading of all the modules? It may be better to include a separate script that imports the other scripts.

hitesh96db: This script is loaded [here](https://github.com/hitesh96db/privly-jetpack/blob/options_interface/lib/privly_ns.js#L8) as a CommonJS module.
Every script needs to load it's own dependencies. Won't there be a cross-dependency(as in circular reference) issue ?

hitesh96db: Also, while loading the module, all the dependencies have to be resolved.

shared/javascripts/context_messenger.js
hitesh96db: I haven't tried it but that should cause the problem you were talking about(i.e, switching out of a tab and deactivating the injection). We have to remove the workers on "pagehide" and "detach" or else it'll cause the same problem you mentioned about [here](https://github.com/privly/privly-jetpack/pull/6#issuecomment-122185645)

smcgregor: OK, I'll classify this as "something to remember for the future"

shared/javascripts/options.js
smcgregor: OK, and the `require` statement won't clear the state of the object when the object is reassigned in that manner?

hitesh96db: State as in ? It's similar to loading a script before another script in an HTML page.

smcgregor: `ls = require("./local_storage.js").ls;` assigns to the global variable `ls`. Assuming the `ls` variable has some property assigned to it already, like `ls.hello = "world";`, then the `require` statement doesn't clobber that property?

hitesh96db: Oh, the property won't be defined.

hitesh96db: But there isn't any problem here with that behavior.

smcgregor: So under this approach if there are `n` script files that all reference each other, there will be `n^2` script inclusions?

hitesh96db: n + (n-1) + (n-2) + (n-3) + ..... + 1 because circular reference is not possible.

smcgregor: OK, `n!`. Still, this seems odd from a performance perspective. Perhaps it would be good to check whether the object is already defined, then `require` it when needed. These variables are in global namespace so they can be shared.

hitesh96db: Well, there won't be anything already defined for a module. A CommonJS module can be thought of a black box, sort of separated from the outside env, or already loaded scripts. Whatever the black box needs, it asks for, and resolves the dependencies.

But there's still a certain level of optimization that can be done. The current way to define the Privly namespace in the background script is  this [file](https://github.com/hitesh96db/privly-jetpack/blob/options_interface/lib/privly_ns.js). But we know that `glyph.js` loads `options.js`, and `options.js` loads `storage.js` and `context_messenger.js`. So, we can make `options.js` export the loaded `storage.js` and `context_messenger.js`, all three of which now becomes available to `glyph.js` and then `glyph.js` can export all three and itself to the background script. So , [this](https://github.com/hitesh96db/privly-jetpack/blob/options_interface/lib/privly_ns.js) file which requires each of them separately can now require all of them using only `glyph.js`. 
I hope that made some sense ?

So we can optimize it at the background script but I doubt we can do anything about the individual modules. It's still a step forward if this works out :)

hitesh96db:
@smcgregor Pushed a new commit.

hitesh96db:
@smcgregor Is this Ok ?

hitesh96db:
Thanks :)

PR #278, opened 2015-08-03
Posting process integration tests

hitesh96db:


test/selenium/configurations.rb
```diff
@@ -36,11 +36,11 @@ def common_configuration_for_firefox_extension
   Capybara.app_host = "chrome://privly"
   @@privly_applications_folder_path = Capybara.app_host + "/content/privly-applications/"
   puts "Packaging the Firefox Extension"
-  system( "cd ../../../../../ && pwd && ./package.sh && cd chrome/content/privly-applications/test/selenium" )
+  system("cd ../../../../../ && pwd && jpm xpi && cd chrome/content/privly-applications/test/selenium")
```
smcgregor: When we submit the extension to Mozilla for distribution, we remove many files that are needed for development but not for the user. `jpm xpi` probably includes everything, so it doesn't mirror our deployment process. Is there a way to make jpm exclude certain files?

test/selenium/configurations.rb
```diff
@@ -36,11 +36,11 @@ def common_configuration_for_firefox_extension
...
 
   # Load the Firefox driver with the extension installed
   @profile = Selenium::WebDriver::Firefox::Profile.new
-  @profile.add_extension("../../../../../PrivlyFirefoxExtension.xpi")
+  @profile.add_extension("../../../../../@privly-0.5.0.xpi")
```
smcgregor: does `jpm xpi` accept parameters to name the file? Having the version number here will require updating this line every time we ship.

test/selenium/specs/ts_all_specs.rb
```diff
@@ -4,4 +4,7 @@
 require_relative "tc_show_injected"
 require_relative "tc_extension_options"
 require_relative "tc_history"
+if @browser == "firefox"
+  require_relative "tc_posting_process"
```
smcgregor: Why only Firefox?

hitesh96db: The tests won't pass for Chrome(because the posting process has changed) and Safari(I just noticed there's a PR open for the posting button).

Should I remove it ?

test/selenium/configurations.rb
hitesh96db: Yeah, there is - [jpmignore](https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/jpmignore).
I'll check it out. The files to ignore --
```
run_test.sh
test.json
*.md
chrome/content/test - Test files
chrome/content/vendor - Jasmine
```
right ?

test/selenium/configurations.rb
hitesh96db: Nope, but I can do something like [this](https://github.com/privly/privly-applications/blob/master/test/selenium/run_all.rb#L58) to find out the version number and then use the correct file name.

hitesh96db: Done.

test/selenium/configurations.rb
hitesh96db: Done - https://github.com/privly/privly-jetpack/pull/9

test/selenium/specs/ts_all_specs.rb
smcgregor: You should add comments to the files indicating why these tests are only run for Firefox.

hitesh96db: Done

test/selenium/specs/tc_posting_process.rb
```diff
@@ -0,0 +1,124 @@
+class TestPostingProcess < Test::Unit::TestCase
+
```
smcgregor: This test also needs a comment indicating it should only be run on Firefox.

hitesh96db: Done

smcgregor:
Merged. I haven't tested the code at this point, but we'll be running the testing system through, errrr...tests in the coming week.

sammyshj:
I tried running the test using,
`ruby run_all.rb -p safari_web -c http://localhost:3000 -r experimental`
and only the posting process test fail. This is because, the test first checks for the privly button however, when the tests are run without installing the extension(by running platform[safari, firefox, chrome]_web), this should not be the case.
The safari_web tests work fine if the posting process test is not run on Safari.
@smcgregor Should I make the changes to exclude Safari from running the test when using safari_web?

hitesh96db:
Oh, I don't think I've tried this with firefox_web.

smcgregor:
Yeah, we need to exclude that test from all the *_web platforms.

PR #279, opened 2015-08-06
Run test cases only if the extension is installed

hitesh96db:
I forgot to add this in the previous PR.

### DIFF
--- a/test/selenium/specs/ts_all_specs.rb
+++ b/test/selenium/specs/ts_all_specs.rb
@@ -1,12 +1,15 @@
 # Put the spec you always want to run here
 require_relative "tc_new"
-require_relative "tc_show"
-require_relative "tc_show_injected"
-require_relative "tc_extension_options"
 require_relative "tc_history"
-# Don't run the posting process tests on Chrome as the
-# posting process has changed.
-if @browser == "firefox" || @browser == "safari"
-  require_relative "tc_posting_process"
+# Run these tests only if the extension is installed
+if @@privly_extension_active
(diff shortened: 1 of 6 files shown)


test/selenium/specs/ts_all_specs.rb
```diff
@@ -6,7 +6,7 @@
 require_relative "tc_history"
 # Don't run the posting process tests on Chrome as the
 # posting process has changed.
-if @browser == "firefox" || @browser == "safari"
+if @browser == "firefox"
```
hitesh96db: Also, I assumed that these same tests would pass for Safari, but there seems to be an issue when these tests run on Safari. @sammyshj will look into it.

sammyshj: Thanks for adding this!

smcgregor:
I think a better fix here is to exclude cases where the extension is not installed.

@sammyshj: Does the test pass for safari_extension?

hitesh96db:
@smcgregor Pushed new changes.

sammyshj:
@smcgregor I am working on getting the safari_extension tests to pass, but, as mentioned earlier, there seems to be a problem with Selenium not being able to load the complete page when navigated to an extension page in Safari. Can you suggest how I should go ahead with this?

smcgregor:
I suspect the Selenium extension is sandboxed from our extension. If they are both packaged into the same extension build, then they should not be sandboxed. I recommend looking into the Selenium code base to assess feasibility.

PR #280, opened 2015-08-13
Bug fix: Storage is dependent on Context Messenger

hitesh96db:
I didn't notice this commit - https://github.com/privly/privly-applications/commit/91e914d0459c28bf94610fcef39552368d9178a8.
Storage is now dependent on context messenger as well. I missed out on that. My bad.

### DIFF
--- a/shared/javascripts/glyph.js
+++ b/shared/javascripts/glyph.js
@@ -49,10 +49,11 @@ if (Privly === undefined) {
 
   // CommonJS Module
   if (typeof module !== "undefined" && module.exports) {
-    module.exports.glyph = Privly.glyph;
     // load dependencies    
     var optionsModule = require("./options.js");
     Privly.options = optionsModule.options;
+    // export interfaces
+    module.exports.glyph = Privly.glyph;
     module.exports.options = optionsModule.options;
     module.exports.storage = optionsModule.storage;
     module.exports.message = optionsModule.message;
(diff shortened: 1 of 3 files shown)


shared/javascripts/glyph.js
```diff
@@ -49,10 +49,11 @@ if (Privly === undefined) {
 
   // CommonJS Module
   if (typeof module !== "undefined" && module.exports) {
-    module.exports.glyph = Privly.glyph;
```
hitesh96db: I've also changed things so that it's now like this --
```
// load dependencies
.. require statements ..
// export interfaces
.. export statements ..
```
or else mixing up the statements can be confusing to look at.

smcgregor:
That is only on the Safari platform since Safari has different storage for the background script and privly-applications. Is this necessary for Firefox?

hitesh96db:
Since the files are common to all, yup.

smcgregor:
But the lines in question are never executed outside Safari. Do you want to include this for stylistic reasons, or will this actually cause an error if not fixed?

hitesh96db:
Privly.message was never used before in `storage`, and Privly.message is used to determine if the platform is Safari, as done [here](https://github.com/privly/privly-applications/commit/91e914d0459c28bf94610fcef39552368d9178a8#diff-e0ec9f60afa3823145d6313f20096fbcR44), so things will break.

smcgregor:
ah, thanks.

PR #284, opened 2015-08-19
Move MVC documentation into priv.ly

SummerWish:


Message/js/messageModel.js
```diff
@@ -6,117 +6,9 @@
  * Fore more details about the Message App, please visit Privly
  * website.
  *
- * About the MVC architecture of the Privly application:
+ * Read documentation of the MVC architecture of the Privly
+ * application here: https://priv.ly/pages/develop.html
```
smcgregor: You should put an anchor on the link to get it to the correct section.

SummerWish: fixed :-)

PR #290, opened 2015-08-23
Inline registration

irdan:
@hitesh96db was able to address all of the issues outlined [here](https://github.com/privly/privly-applications/pull/180#issuecomment-78156797) a few months ago.  I merged the changes into gsoc-develop and updated the name of a local storage key.

PR #292, opened 2015-08-25
Add inline registration

smcgregor:
I added a few changes to #290. @irdan can you give my changes a once over?


irdan:
@smcgregor HTML should in fact live in HTML instead of javascript, good call.  The styling is also way better than the vanilla I had in place.

Not that most people would do this, but you get a weird UI if you do the following:

* Submit a valid email 
* Disconnect your network
* Submit another request

This causes both the success and the failure message to appear at the same time because the success does not clear when the failure is presented.  If you submit another valid email, the error message disappears but there is no 'new' success messaged added.

I also noticed that if you close the modal and then enter it again the state is maintained.  Which seems a bit odd. If you care to address the issues I mentioned, I made a [gist](https://gist.github.com/irdan/edb7183b10c2f88ca5ee) of the diff that would address them.

smcgregor:
Updated the pull request per your comments.

irdan:
Awesome :+1: to merging this.

PR #305, opened 2015-11-06
Move SauceLabs selenium firefox version to 38

smcgregor:
This fixes JetPack's Selenium tests for me.


smcgregor:
(without this change the browser cannot navigate to the extension context pages)

PR #307, opened 2015-12-12
Improved Metadata Design

MaxBorsch:
Improved overall look of the msg show page.

- mborsch (GCI)

narrow
wide

### DIFF
--- a/shared/css/top/top.css
+++ b/shared/css/top/top.css
@@ -38,11 +38,61 @@ body {
   word-wrap: break-word;
 }
 
-#side_block {
-  background:white;
-  border-radius:6px;
-  padding:13px;
+#post_content {
+    background: white;
+    border-radius: 6px;
+    padding: 13px;
+    box-shadow: 4px 5px 0px #8E8980;
 }
+
+#post_content p {
+  margin: 0;
+  word-wrap: break-word;
+}
+
+#cleartext {
+  height: 100%;
+}
+
+@media(min-width: 992px){
+  #dashed_hr {
+    display: none;
+  }
+}
+
+#dashed_hr {
+  border-top: 3px solid #CCBEA8;
+  width: 80%;
+  margin-bottom: 50px;
+  border-style: dashed;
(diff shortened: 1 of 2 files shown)


irdan:
For the purposes of feedback and iteration, would you mind submitting a screen shot of your changes?

MaxBorsch:
@irdan Oh yes I was going to do that, forgot because I already submitted the screenshots on GCI.

smcgregor:
I think this looks improved and is implemented almost entirely in CSS. If it looks good to @irdan, I'll look over the CSS and merge.

@irdan He edited the pull request, but I don't think it will have emailed you.

irdan:
LGTM! :+1:

shared/css/top/top.css
```diff
@@ -38,11 +38,61 @@ body {
...
+  margin: 0;
+  word-wrap: break-word;
+}
+
+#cleartext {
+  height: 100%;
+}
+
+@media(min-width: 992px){
+  #dashed_hr {
```
smcgregor: I'll run this on my computer, but I think this should be tied to a responsive value from [bootstrap](http://getbootstrap.com/). I'm not sure what that is at the moment.

@spectralsun do you know how to trigger this CSS rule when bootstrap makes the columns stack vertical?

MaxBorsch: Uh yeah I wasn't sure how to trigger it, so I found that rule in bootstrap. 
```css
@media(min-width: 992px){
```
It's the exact same rule that bootstrap uses to when converting to vertical stacked, so there shouldn't be any problems other than it not being the cleanest approach.

smcgregor: That is hackier (on Bootstrap's account) then I would have expected since different devices have different pixel densities. Good to know, thanks.

smcgregor:
Merged to the develop branch. Future pull requests should be made against that branch. Great work!

https://github.com/privly/privly-applications/commits/develop

PR #308, opened 2015-12-17
Re-styled Metadata

MaxBorsch:
It's me again :D

Updated metadata section. -mborsch GCI

### DIFF
--- a/shared/css/top/top.css
+++ b/shared/css/top/top.css
@@ -98,11 +98,106 @@ body {
   border-bottom:1px solid #242424;
 }
 
+#side_block .glyphicon-plus {
+  color: #B3FFAF;
+  margin-right: 5px;
+}
+
+#side_block .glyphicon-minus {
+  color: #FFAFAF;
+  margin-right: 5px;
+}
+
+#side_block .glyphicon-pencil {
+  color: #FFFEAF;
+  margin-right: 5px;
+}
+
+#side_block .destroy-button {
+  width: 90%;
+  margin-left: 5%;
+  background: #FFAFAF;
+  border: dashed;
+  border-color: #FB8C8C;
+  font-weight: bold;
+  color: #484848;
+  margin-bottom: 10px;
+  transition: all 0.5s;
+}
+
+#side_block .destroy-button:hover {
+  background-color: #FF6161;
+}
+
+#side_block .edit-button {
+  width: 90%;
+  margin-left: 5%;
+  background: #FFFEAF;
+  border: dashed;
+  border-color: #D4D266;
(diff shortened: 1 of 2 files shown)


smcgregor:
Screen capture:

screen shot 2015-12-17 at 8 10 36 am

smcgregor:
What it looked like 2 weeks ago:

mxnpman

MaxBorsch:
I forgot to mention, the destroy and update buttons currently link to "/", I wasn't sure where they should take you when clicked and didn't want to mess with code. Making them work could be another task. Also note that another task should be changing the metadata to fit with the new headings and be more formal, for example notice how it says data source twice, because the actual metadata includes "Data source:"

smcgregor:
Hello Again!

* The layout shifts between header, body, and button too many times in a small area which makes it look a bit cluttered. Maybe put each section of the metadata into its own darker patch, including the header?
* I'll hook the destroy and update buttons into Javascript so they work.
* Put the "source data" section next to the "source link" section.
* Put the "update" section above the destruction section. Retitle the section to "Edit Content"
* When the user doesn't have permissions on the post, it doesn't display the proper sections, see below.

screen shot 2015-12-17 at 8 21 15 am

MaxBorsch:
K, will do once I come back from midterm exam.

MaxBorsch:
@smcgregor Updated. How's this?
new

smcgregor:
Getting closer. I still think it is a bit visually confusing.

> Salience is the state or condition of being prominent. The Oxford English Dictionary defines salience as "most noticeable or important."

Particularly, the use of the darker area for both dividing blocks and highlighting the contents of the block make the divisions between blocks less salient. Perhaps it would be better to eliminate the darker area for "Created around", "automatically destroyed around", and "https://". These elements don't need to be highlighted if the user can easily find their section.

MaxBorsch:
@smcgregor Updated, removed black boxes around text.

smcgregor:
@irdan @sammyshj @spectralsun Any input around this design? I am going to do the JS on the buttons and likely remove the actions from the nav bar.

MaxBorsch:
I agree with removing the actions from the Navbar, they're too unnoticeable since they're tucked away in a dropdown. It would make the Navbar less cluttered as well.

sammyshj:
@smcgregor This looks a lot better than what it previously used to be.
A small change that could be done is to have the height of the `Source Link` textbox to fit the height of the URL. Otherwise, it looks great!

smcgregor:
@sammyshj I made the source link select its entire content upon being clicked.

Thanks all!

PR #311, opened 2015-12-27
Update base64.js

vg3095:
Just improving the code based on the recommendations found on code climate.
Replaced '==' with '==='.
It will ensure to check equality of both number and type


smcgregor:
Thanks for working for conformance, but we didn't develop this library so we only make changes if something is broken or insecure.

PR #313, opened 2015-12-29
build.py file is now compatible with both Python2 and Python3

opg7371:
issue#312 can be closed now
This build.py file is now compatible with both python2 and python3


smcgregor:
* Pull requests shouldn't have the built HTML (this makes it easier for me)
* The built HTML is not encoded properly. You can see the newline characters rendered as `\n`.
* You should call the make_pretty function even if the HTML is encoded as bytes. You will need to convert the bytes for this to work, which should also solve your encoding issue.

opg7371:
Hi sean please check it out.
I fixed the newline problem too...

build.py
```diff
@@ -67,10 +67,25 @@ def make_readable(html):
   # Beautiful soup breaks textarea formatting
   # since it adds extra whitespace. If you use "pre"
   # tags, you should be warry of the same problem
-  return re.sub(r'[\ \n]{2,}</textarea>',
+  # To check the type of prettyHTML 
+  #print(type(prettyHTML))
```
smcgregor: Some leftover code?

build.py
```diff
@@ -170,6 +185,7 @@ def get_link_creation_apps():
     if "manifest.json" in filenames:
       f = open(dirname + "/manifest.json", 'r')
       template_list = json.load(f)
+
```
smcgregor: nit pick: no need to add a line here.

build.py
```diff
@@ -108,7 +123,7 @@ def get_link_creation_apps():
   creation_apps = []
   for dirname, dirnames, filenames in os.walk('.'):
     if "manifest.json" in filenames:
-      f = open(dirname + "/manifest.json", 'r')
+      f = open(dirname + "/manifest.json", 'r' )
```
smcgregor: nit pick: no need to add a space here.

build.py
```diff
@@ -67,10 +67,25 @@ def make_readable(html):
...
+  if(type(prettyHTML)==str):
+    #For python2 the type is String
+    #print(type(prettyHTML)) 
+    return re.sub(r'[\ \n]{2,}</textarea>',
                "</textarea>",
                prettyHTML)
-
+  else:
+    #For python3 the type is bytes
+    #print(type(prettyHTML))
```
smcgregor: remove debugging print statements

opg7371:
can you tell me where exactly the work is needed..?

smcgregor:
Summarize the changes you made so I don't need to look through a diff to figure out what changed. I'll look through a diff again when you have addressed the earlier comments.

opg7371:
To be clear.i need to summarize the changes i did in that build.py file itself as comments on top or some specfic position or some where else...?

smcgregor:
In the comment box of GitHub. It emails me and gives me more details about the changes.

opg7371:
In the **build.py** file ,only the line number 70-80 are changed.
the same lines are ** #The if statement is needed to maintain support for both python2 and python3	
  if(type(prettyHTML)==str):
    #For python2 the type is String
    return re.sub(r'[\ \n]{2,}</textarea>',
               "</textarea>",
               prettyHTML)
  else:
    #For python3 the type is bytes
    return re.sub(r'[\ \n]{2,}</textarea>',
               "</textarea>",
               prettyHTML.decode('utf-8'))**

smcgregor:
I see why you were confused now. I am looking for a summary of the changes you made to the original pull request, not a summary of the pull request as it currently is. This makes it faster for me to re-review the pull request.

opg7371:
The changes that i made in the build.py for privly applications is made in from line 70 to 72. The problem was that the type of prettyHTML came out to be different in python2(str) and python3(bytes).So i committed the changes can be seen in line number 71 to 81.The lines changed can be seen in the above commit.

PlainPost/js/show.js
```diff
@@ -56,10 +58,11 @@ function processResponseContent(response) {
 
   // Make all text areas auto resize to show all their contents
   if ( ! privlyHostPage.isInjected() ) {
+   console.log("working correctly till here");
```
irdan: nit: remove debugging console statements

PlainPost/js/show.js
```diff
@@ -56,10 +58,11 @@ function processResponseContent(response) {
...
     $('textarea').autosize();
   }
-  privlyHostPage.resizeToWrapper();
-}
+   privlyHostPage.resizeToWrapper();
```
irdan: nit: unneeded whitespace change

build.py
```diff
@@ -67,10 +67,19 @@ def make_readable(html):
...
+    return re.sub(r'[\ \n]{2,}</textarea>',
                "</textarea>",
                prettyHTML)
-
+  else:
+    #For python3 the type is bytes
+    return re.sub(r'[\ \n]{2,}</textarea>',
+               "</textarea>",
+               prettyHTML.decode('utf-8'))
+
```
irdan: Having a single return statement would make this more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).  One way to refactor with that aim would be conditionally decoding `prettyHTML` before it is used in the return statement.

opg7371: Yes,i did that at the beginning .however it was producing some /n characters in the HTML pages meaning it was not decoded properly.So i approached this way

smcgregor: I think his suggestion is to assign `prettyHTML` to `prettyHTML.decode('utf-8')` in the `if` statement and have a single return.

smcgregor:
Your pull requests are getting mixed. You should be doing all your work on feature branches. Since you are working on two pull requests at the moment, the two branches should probably be `feature-build-python-3` and `feature-improve-footer`. These branches should be current with the `develop` branch and should not share any commits between them that are not in `develop`. Also, your pull request should request to be merged into `develop`, not master.

opg7371:
Closing this issue
This issue is reopened at #349

PR #320, opened 2016-01-14
Feature build python 3

opg7371:
he changes that i made in the build.py for privly applications is made in from line 70 to 72. The problem was that the type of prettyHTML came out to be different in python2(str) and python3(bytes).So i committed the changes can be seen in line number 70 to 79.The lines changed can be seen in the above commit.


opg7371:
Just one thing,I added and committed only the build.py file and when i created a pull request,All the files (html) are here.Am i missing something sir.

irdan:
If I recall correctly @smcgregor would prefer this PR target the privly-applications develop branch.  Also, it appears that more than just the two commits you describe were included in this PR.

Issue #124, opened 2014-11-16
Console errors in show.js when there is no content to show


irdan:
Line 59 of `show.js`, which is `$('textarea').autosize();`, currently throws `Uncaught TypeError: undefined is not a function`  when there is not content displayed on the page.  This shouldn't happen.

To reproduce this error, just directly access `privly-applications/PlainPost/show.html`.  Note that there is a race condition and the page may need to be refereshed a few times in order to reproduce.

opg7371:
The **$('textarea').autosize();** is a old way to  write ..the new way to write is **autosize($('textarea'))**. But after fixing it ,it shows a reference error "autosize is not defined".I checked out further and it says that npm should be installed.How do i check it?
I want to fix this issue.Please assign me this issue.
the link for the same is: http://www.jacklmoore.com/autosize/

smcgregor:
We have the older version of the script, so the new way will not be defined.

opg7371:
so then if we are using the older version of the script,am i allowed to update it or should i do something with the older version

smcgregor:
You can update it if you want, but I would prefer to keep the older version unless there is a security or functionality issue with it.

opg7371:
I tried but the reference error is not fixing..can we chat on irc for a moment

smcgregor:
If I recall correctly, the problem with this ended up being a timing issue with the loading of the autosize library. When there is no content to show, the server returns the content almost immediately and the autosize function is called. However, if the autosize function is called too quickly, the library will not be loaded because the script is conditionally included.

When solving a race condition, it is usually best to figure out a way to eliminate the race condition rather than hack a solution on top of it. I think an acceptable solution that doesn't require larger changes would be to call the autosize function after the user clicks on the text area, rather than immediately, but this would also require some testing.

PR #325, opened 2016-02-06
Fixed the console error in the show.html file

opg7371:
This is regarding the console.log error in the show.html file in Plainpost section
Issue number #124


opg7371:
The file autosize.js was not included properly thus giving the autosize error.It was initailly called using meta but as soon as i called it using script tag the console.log error was gone and the page was working fine in my case.
Looking forward for suggestion

smcgregor:
We don't want to include this library by default because it will be loaded every time a privly app is injected into a page. It sounds like the problem may be the autosize library not being loaded before it is called. I would prefer to solve this problem by ensuring the library is loaded before attempting to initialize it, rather than including it in a different manner.

Issue #328, opened 2016-02-15
Exclude Files from CodeClimate


smcgregor:
This is intended to be an introductory issue.

Our [CodeClimate report](https://codeclimate.com/github/privly/privly-applications) gives us a failing grade because many files that should not be graded are included in the report. The way we exclude these files is with a [.codeclimate.yml](https://docs.codeclimate.com/docs/example-codeclimateyml) file placed in the root directory.

Files to exclude
- All scripts in a `vendor` directory
- All ZeroBin scripts
- All test-related scripts

theaverageguy:
Any part of it still open?

smcgregor:
@yash14123 Still resolving a PR on it (#336), but it is otherwise nearly closed.

PR #331, opened 2016-02-18
Add .codeclimate.yml

ayushgp:
Added .codeclimate.yml to stop Code climate from tracking tests, ZeroBin scripts and vendor scripts.

PR #333, opened 2016-02-18
Improved the code to be excluded in codeclimate.

opg7371:
This is regarding the issue #328 
I have excluded all the files which are-
-All scripts in a vendor directory
-All ZeroBin scripts
-All test-related scripts which are 
           -test files in test directory
           -test files in other directories

please review it too....:)

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,21 @@
+engines:
+	rubocop:
+		enabled: true
+	golint:
+		enabled: treu
+	eslint:
+		enabled: true
+	csslint:
+		enabled: true	
+ratings:
+	paths:
+	-app/**
+	-lib/**
+	-"**.rb"
+	-"**.go"
+exclude_paths:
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -0,0 +1,21 @@
...
+	golint:
+		enabled: treu
+	eslint:
+		enabled: true
+	csslint:
+		enabled: true	
+ratings:
+	paths:
+	-app/**
+	-lib/**
```
smcgregor, line 13: It looks like you copied these directories from an example. We have no `lib` directory, `app` directory, or `.go` files. We also don't want to rate the Ruby files in this repo.

smcgregor:
Pull requests should be made against the `develop` branch.

#331 was opened earlier and has fewer issues. I am going to merge #331, but there is still room for improvement in it. Specifically, your list of excluded files is more complete.

PR #342, opened 2016-02-21
Updated the .codeclimate.yml file

opg7371:
Removed the tests directory as you recommended.
Updated the .codeclimate.yml file
Please review ...:)
#328

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,4 +1,5 @@
 exclude_paths:
 - vendor/**/*
 - ZeroBin/**/*
-- tests/**/* 
+- test/*/*
+- */test/ 


.codeclimate.yml
```diff
@@ -1,4 +1,5 @@
 exclude_paths:
 - vendor/**/*
 - ZeroBin/**/*
-- tests/**/*
```
smcgregor, line 4: What is the difference here between two `**` and one `*`?

smcgregor:
I think this is your third pull request making similar changes to the codeclimate file. You should be updating a single pull request, or at least closing the pull requests you open that are no longer valid.

opg7371:
yes i will close other pull requests......I created this since you hinted me to remove a line which was not needed.Is there anything else that i should add ??
Should i reference this to the pull request that is still open??

.codeclimate.yml
opg7371, line 4: I don't know exactly but if have to guess,then they both are pretty same.I will update them

smcgregor, line 4: Please don't guess. You can't ask for something to be merged if you don't know what the code does.

opg7371, line 4: I researched on this thing...they (\* & *\* ) are both same and does same work in this case for sure

opg7371:
The new pull request is present at....#355
So i am closing this issue

.codeclimate.yml
smcgregor, line 4: OK, thanks.

Issue #155, opened 2015-02-23
Fix CodeClimate Issues


smcgregor:
This is an entry level issue for users to become familiar with the Privly Project's code base. You should pick a file from [CodeClimate](https://codeclimate.com/github/privly/privly-applications) and improve files per the recommendation of the programmatic code quality system.

Do not close this issue in your commits since this is intended to be a more permanent issue.

maniksingh92:
CodeClimate issue pull request #208

sudheesh001:
@smcgregor Since this issue is merged in #182  I think its good to close this issue.

smcgregor:
This repository still needs improvement, according to CodeClimate. I will leave the issue open until CodeClimate stops complaining since it is a great entry-level issue.

See: https://codeclimate.com/github/privly/privly-applications

PR #343, opened 2016-02-21
Some program style changes (linting and declarations)

goelakash:
#155


smcgregor:
These look like good changes, except we don't want any changes to scripts in the vendor directory. If we change those scripts, we will fork the originating code and move it out of the vendor directory. Linting changes are not sufficient reason to fork someone else's code.

Also, it looks like you pulled in the merge commit from master. It is best to fork from the branch you plan on generating a pull request for.

goelakash:
So should I close this PR, since I would need to branch from develop and stay away from vendor scripts except for some critical changes?

smcgregor:
Sure, if you want to do it that way. You should include a reference to this pull request in the new one's description.

PR #345, opened 2016-02-21
Edited code after unit tests from karma setup

goelakash:


smcgregor:
We need more description than this. What is the problem this is addressing? How do I recreate the problem? What files are changed, and why? etc.

goelakash:
Closing PR because I can't recreate the errors.

PR #348, opened 2016-02-21
Added new unit tests for event_emitter.js

goelakash:
For #157, I made some small unit tests to make sure an undefined event was not getting passed to the function.

### DIFF
--- a/shared/javascripts/event_emitter.js
+++ b/shared/javascripts/event_emitter.js
@@ -25,38 +25,44 @@ if (Privly === undefined) {
 
   /**
    * Add an event listener.
-   * 
+   *
    * @param  {String}   ev Event name
    * @param  {Function} listener
    */
   Privly.EventEmitter.prototype.on = function (ev, listener) {
+    if(ev === undefined) {
+      return false;
+    }
     if (this.__listeners === undefined) {
       this.__listeners = {};
     }
     if (!this.__listeners.hasOwnProperty(ev)) {
       this.__listeners[ev] = [];
     }
-    if (this.__listeners[ev].indexOf(listener) === -1) {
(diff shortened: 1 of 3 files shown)


shared/test/context_messenger.js
```diff
@@ -10,7 +10,7 @@ describe ("Messaging Test Suite", function() {
   it("can send a URL to the extension", function() {
     var message = {"handler": "privlyUrl"};
     Privly.message.messageExtension(message);
-  });
```
smcgregor, line 13: Accidental indent.

shared/test/context_messenger.js
```diff
@@ -97,7 +97,7 @@ describe ("Messaging Test Suite", function() {
 
 describe('Privly.message.adapter.*', function () {
 
-  it('can remove listeners', function (done) {
+  it('can add and remove listeners', function (done) {
```
smcgregor, line 100: Good clarification, but it would be better to have this in a separate commit since it is not related to adding event emitter test coverage.

shared/javascripts/event_emitter.js
```diff
@@ -25,40 +25,46 @@ if (Privly === undefined) {
...
 
   /**
    * Remove an event listener.
-   * 
+   *
    * @param  {String} ev Event name
    * @param  {Function} listener
    */
   Privly.EventEmitter.prototype.off = function (ev, listener) {
+    if(ev === undefined) {
```
smcgregor, line 54: You are changing the signature of the function to make it easier to test. This is generally fine, but at this point the two potential return values are `false` and `undefined`, which can introduce problems in the boolean interpretation of the return value. It might be better to throw an error in the event of invalid input, and test for the error from the unit test.

PR #349, opened 2016-02-21
Updated the build.py to be compatible with python3

opg7371:
The changes that i made in the build.py for privly applications is made in from line 70 to 72. The problem was that the type of prettyHTML came out to be different in python2(str) and python3(bytes).So i committed the changes can be seen in line number 78 to 86.The lines changed can be seen in the above commit.

### DIFF
--- a/build.py
+++ b/build.py
@@ -75,10 +75,15 @@ def make_readable(html):
   # Beautiful soup breaks textarea formatting
   # since it adds extra whitespace. If you use "pre"
   # tags, you should be warry of the same problem
-  return re.sub(r'[\ \n]{2,}</textarea>',
-               "</textarea>",
-               prettyHTML)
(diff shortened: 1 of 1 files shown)


opg7371:
This branch has some comments which might help 
#313

build.py
```diff
@@ -75,10 +75,15 @@ def make_readable(html):
   # Beautiful soup breaks textarea formatting
   # since it adds extra whitespace. If you use "pre"
   # tags, you should be warry of the same problem
-  return re.sub(r'[\ \n]{2,}</textarea>',
-               "</textarea>",
-               prettyHTML)
+  #The if statement is needed to maintain support for both python2 and python3	
+  if(type(prettyHTML)!=str):
+  	#For python2 the type is String and for python3 the type is bytes
```
smcgregor: nit: the whitespace is inconsistent here

Issue #317, opened 2016-01-09
Increase font size of cleartext content when not injected.


smcgregor:
According to [this blog post](https://sumedhsupe.wordpress.com/2016/01/09/my-first-go-with-privly/), the cleartext is a bit small relative to the rest of the application. We should increase the text size for the cleartext in [top.css](https://github.com/privly/privly-applications/blob/master/shared/css/top/top.css).

PR #337, opened 2016-02-19
Updated the cleartext element

opg7371:
Updated the size of cleartext element in the top.css file
Please review .....:)
#317

PR #354, opened 2016-02-25
Updated the cleartext element

opg7371:
Updated the size of cleartext element in the top.css file
Please review .....:)
#317


opg7371:
#337

opg7371:
It looks like this Finally....
cleartext-min
content-large

PR #355, opened 2016-02-25
Updated the .codeclimate.yml file

opg7371:
Updated the .codeclimate.yml file
Please review ...:)
#328

The previous data can be found here....
#342


opg7371:
they (\* & **) are both same and does same work in this case for sure.

Issue #304, opened 2015-11-06
Fix page title of extension options


irdan:
Right now in Chrome the page title of the extension options page is `New Options`.  This should be updated to a more meaningful title such as `Options`.

opg7371:
can you provide a screenshot for what you want to say..i see there is no term as New Options

prastut:
This issue is present on most pages. Added a global fix in PR #360

prastut:
@smcgregor You can close this issue now.

PR #360, opened 2016-02-27
Options and History page title bugfix ( #304)

prastut:
Added venv/ to gitignore for vitualenv wrapper.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@
 test/sauce.json
 test/node_modules/*
 test/coverage
-venv/
+
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 test/sauce.json
 test/node_modules/*
 test/coverage
+venv/
--- a/templates/new.html.template
+++ b/templates/new.html.template
@@ -1,7 +1,7 @@
 {% extends "templates/base.html.template" %}
 {% block head %}
 
-    {% block title %}<title> {{ name }}</title>{% endblock %}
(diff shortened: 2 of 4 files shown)


templates/new.html.template
```diff
@@ -1,7 +1,7 @@
 {% extends "templates/base.html.template" %}
 {% block head %}
 
-    {% block title %}<title> New {{ name }}</title>{% endblock %}
+    {% block title %}<title> {{ name }}</title>{% endblock %}
```
smcgregor: This removes "New" from all apps using `new.html.template`. A better solution is to implement the title block in the options subtemplate, which will take precedence over the title in this file.

.gitignore
```diff
@@ -5,3 +5,4 @@
 test/sauce.json
 test/node_modules/*
 test/coverage
+venv/
```
smcgregor: This is not related to the issue you are closing, so it should be in a separate pull request.

smcgregor:
Close, would you like tips on Jinja2, or did my inline comments point you in a good direction?

prastut:
Inline comments did the job. How do I proceed further? Make the necessary changes in the same branch or open a new branch and then do a PR?

smcgregor:
Push commits to your same branch. When it is ready for merge I'll ask you to squash your commits so we don't pollute the history.

prastut:
Documenting new commits:
- Removed venv from gitignore. Will make a seperate PR.
- Options page and History page get there own blockquote.

Issue #152, opened 2015-02-21
Prevent re-authentication on Android Platform.


vshivam:
Whenever any of the applications are loaded, they verify the auth_token with the server. In the Android application, this looks like a bit of overhead because every time the user switches options, the token is verified. Can we skip this step when on the android platform ?

smcgregor:
I think the problem is that mobile has a higher latency so it is far more noticeable. I don't think it is necessarily the best option since I'd like to avoid the check on the browser extension as well, but I agree that skipping the step on Android is reasonable for the time being.

mkash32:
I was thinking of modifying the `initPrivlyService` method in network_service.js. Before the ajax request is sent we keep a check to see if the platform is android and if it has been authenticated before. If it is satisfied then we can directly invoke the `loggedInCallback`. I tried this out and it worked out but it does have a few kinks. Namely, the ajax request loading image is being displayed (could be removed with a simple `$('#loadingDiv').hide();` call) and there is some lag between pressing save and the url being displayed (haven't looked into it yet). 

Is this an appropriate approach or should I try something else?

smcgregor:
@mkash32 There are subtle timing issues here, since the user won't be able to create content until they have the CSRF token that is returned by this request. I am trying to figure out how to balance expediency of showing the user the content creation forms by default, and the case where the user is not logged in and should be redirected immediately. I don't think adding more case statements for Android is a great solution because it increases complexity where we should just try to mitigate slow connections somehow, maybe by presenting all the form elements by default, but disabling the "submit" buttons until the user is authenticated.

mkash32:
@smcgregor I like the solution you suggested, it would be better for all of the other platforms also. So basically, the user will be able to type in all the content while the authentication is being done in the background. I think the time taken to type a message of average size would be a little more than that of authentication (given a reasonable Internet connection speed). By the time the user is done typing their content, the "Save" button will be enabled (in most cases). It would probably be good to keep some kind of loading indicator next to the save button to communicate to the user that authentication is being done. We could use the same ajax loading gif and place it next to the button. 

Frankly, I wasn't that satisfied with the solution I suggested because it was too obtrusive to the flow of privly-applications (therefore a possible source of more bugs), but I thought we were just trying to target the android platform for now, so I wasn't able to find a better option. I'll broaden my solution approach from next time.

PR #363, opened 2016-02-28
Changed Authentication Flow

mkash32:
#152

### DIFF
--- a/Message/new.html
+++ b/Message/new.html
@@ -193,9 +193,6 @@
         </span>
         Create New Message Link
        </h2>
-       <p id="loadingDiv">
-        <img alt="loading" src="../shared/images/ajax-loader.gif"/>
-       </p>
        <label for="content">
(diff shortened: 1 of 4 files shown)


irdan:
I have not tested this yet.  However a few initial comments:
-  Do not commit built html, instead only modify the templates.
-  Typo, whitespace, or any change not related to fixing #152 should be made in a separate PR or at the very least in a separate commit.

mkash32:
@irdan I didn't understand the first comment about modifying templates. In this situation where I want to make changes to PlainPost/new.html and Messages/new.html which templates should I modify?

irdan:
You can find all of the templates and subtemplates in the repo by running `find . | grep template$`.  After you make the changes to your desired templates, you will want to run the [build system](https://priv.ly/pages/develop#BuildSystem) to verify your changes appear in the static html as expected.

Issue #358, opened 2016-02-26
The History Application is supremely broken for a Mobile Layout. 


vshivam:
- Some screenshots from chrome using the device mode. It'd be great if everything fits on the screen _and_ looks good. 
- ![screenshot from 2016-02-26 15-35-38](https://cloud.githubusercontent.com/assets/831071/13346155/367039e4-dc9f-11e5-9bd6-cf3bb6d83b25.png)
- ![screenshot from 2016-02-26 15-43-16](https://cloud.githubusercontent.com/assets/831071/13346230/c90d82de-dc9f-11e5-96ea-cc3d1a341cb3.png)

prastut:
True.
Taking this up after my PR get's approved.

vshivam:
Sounds good. It'd be great if you could rethink the approach rather than just trying to _fix_ the existing one. I believe we can do much better.

prastut:
If we are going there, I feel the whole website could be rethought in that perspective. I was talking with @irdan on IRC about this only, and I did a get a good grasp about the vision Privly is trying to build. I am a big fan of Don Norman's work in cognitive science, and I feel we could bring in more user-centric elements too. Privly is a lot more than _just_ being an extension. I just feel it needs to convey that too.

We could discuss further if you guys agree too.

opg7371:
![histroy_chrom](https://cloud.githubusercontent.com/assets/10517011/13352602/509e758c-dcb4-11e5-90ee-7654bb866750.png)

Looks cool on my chrome...atleast..it is resizeing properly
am i missing something...@vshivam
can you tell me how to generate it @vshivam

prastut:
@opg7371 We need to redesign the approach taken by the History page. We could collaborate on this if you want?  I would love to brainstorm the design process with you. 

The table can be rethought. Some suggestions:

![screen_shot_2016-02-01_at_8 35 19_pm](https://cloud.githubusercontent.com/assets/10279686/13360779/cf8bb0e0-dce0-11e5-9c18-94c204d526dc.png)

Sorry for the long screenshot but I wanted to make my point noticed in the context where we have a lot of entries.
- I really love YouTube's history page. We can divide the page into 2 sections. 
  - One for the preview. Reduce clicks by:
    - Doing away with the modal that get's triggered. The options present on the modal can be brought in the form of dropdown only. 
      - One issue I can think of is the length of the message being posted. Would have to make it a scroll-able content -> than modal seems a better option.
[...]

PR #364, opened 2016-02-29
Develop fix histroy content

opg7371:
For now the history table is fully resizeing according to the mobile view...Also the pop-up is also resizeing. 
@smcgregor please review...:)
#358

### DIFF
--- a/History/new.html.subtemplate
+++ b/History/new.html.subtemplate
@@ -59,37 +59,44 @@
 {% endblock %}
 
 {% block form %}
- <div class="col-md-12" id="table_col">
-    <h1>History</h1>
-    <p id="loadingDiv">
-      <img src="../shared/images/ajax-loader.gif" alt="loading..." />
-    </p>
-    <table class="display" id="posts" width="100%">
-      <thead>
-        <tr>
-          <th>Preview</th>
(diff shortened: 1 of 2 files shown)


smcgregor:
Can you post a screen capture with the width being cell phone sized?

opg7371:
Currently looking like this...
in the chrome by changing to device mode
histroy
in my own mobile device
screenshot_2016-03-01-02-09-35_ly priv mobile

screenshot_2016-03-01-02-12-25_ly priv mobile

opg7371:
sorry for the second and third screenshot ....they are from my mobile device...i didn't thought that they will look like this.....posting small one ....

History/new.html.subtemplate
```diff
@@ -59,37 +59,44 @@
...
-        <tr>
-          <th>Preview</th>
-          <th>Created</th>
-          <th>Burnt After</th>
-          <th>Updated</th>
-        </tr>
-      </tfoot>
-    </table>
-  </div><!-- /.col-md-12 -->
+ <div class="row">
```
smcgregor, line 62: I think this might already be wrapped in an enclosing div with class "row".

